PR #26, opened 2014-09-10
Item "toString()" and PWX "macro" implementation

d00t3y:
Itemdata:

-Added "ToString()" Method to Item-class -> Needs further testing for different itemtypes/variatians. It's supposed to return the same Syntax as "Ingame ctrl+c".
-Added "Copy Itemdata"-button to "SetBuyoutView" -> copy itemdata to clipboard
-moved some objects to "Item" (eg. from Gear) because they are used/necessary for "toString()" -> seems to work fine (since gear inherits item)

PWX/Pricesearch:

-Added Project for PWX-Api
-Implemented simple HTTP-request for the "macro"-URL (should be changed to json, as soon as the api from pwx is online)
-Added "Get Price" to "SetBuyoutView" --> Shows the result as Tooltip on Button

Issues:
Background of tooltip for GetPrice is always transparent. Maybe an issue with another style I cannot find, but tried background, opacity, opacity mask,... Using only border (without tooltip, like search checkbox) could fix the background issue, but then the tooltip cannot be triggered from codebehind...

Bugreports and feedback are always appreciated ;)

### DIFF
--- a/POEApi.Model/Gear.cs
+++ b/POEApi.Model/Gear.cs
@@ -7,22 +7,21 @@ namespace POEApi.Model
     {
         public Rarity Rarity { get; private set; }
         public List<string> FlavourText { get; set; }
-        public List<Socket> Sockets { get; set; }
+        //public List<Socket> Sockets { get; set; }
         public List<Gem> SocketedItems { get; set; }
         public List<string> Implicitmods { get; set; }
-        public List<Requirement> Requirements { get; set; }
-        public GearType GearType { get; set; }
+        //public List<Requirement> Requirements { get; set; }
+        //public GearType GearType { get; set; }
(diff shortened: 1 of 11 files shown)


Stickymaddness:
Guys I'm going to close this pull request, since I just haven't got to it and the repo has changed considerably. It's still something I would love to add.

trackpete:
No worries - my API is very different now. I have actually been working on a more advanced setup with an ES back end that allows much better analysis, however I may end up having to retire my indexer. Starting with the latest leagues, a huge amount of the playerbase has switched to Acquisition, which no longer bumps the forum threads but instead just sends a trigger to poe.trade on the back-end. Unless I can get the developer to change this behavior, it will effectively kill every other indexer.

PR #33, opened 2014-11-05
Adding support for automatic refresh of online status in poe.trade

fer662:
I added a small feature that helps the user keep his user online in poe.trade/poe.xyz for as long as he keeps Procurement open. Hoping you'll like it and merge it.

Disclaimer: my c# might be a bit rusty, let me know if you need me to clean something up

### DIFF
--- a/POEApi.Model/Settings.cs
+++ b/POEApi.Model/Settings.cs
@@ -89,7 +89,13 @@ namespace POEApi.Model
 
         private static ShopSetting createShopSetting(XElement shop)
         {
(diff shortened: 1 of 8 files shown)


Stickymaddness:
I've finally had a chance to look into this, I'm curious why you have a trade url per league? From what I remember when you pm the xyz bot it only gives you one unique url, regardless of league. 

Unless there is something I'm missing, I think the url can be a single setting, independent of the shop settings. I'm also wondering about the polling starting as soon as Procurement is started, what if the user logs in offline mode.

Other than that, it looks good and I want to add this to Procurement

PR #58, opened 2014-12-26
Added "Advanced Search" filtering

omgftw:
The advanced search section is convenient, but if you're looking for specific things, it can be difficult to search through the list manually. Simply added a textbox in the "advanced search" tab where the user can type text. On KeyUp it goes through all the categories and sets the visibility attribute to "Collapsed", via a binding, if it doesn't partially match the user's input.

Filter with nothing typed: http://i.imgur.com/RsFqWSl.png
Filter with "resist" typed: http://i.imgur.com/Tk71RKU.png


Stickymaddness:
This is great, the advanced search has been a bit unwieldy for a while. Thanks for the contribution!

PR #59, opened 2014-12-26
Added "Advanced Search" filters for all primary gem colors and amounts

omgftw:
Added Advanced Search Filters for every amount of the three primary gem colors.

This allows you to easily find gear in your stash for specific gem setups.

Example (A setup requiring RRBG):
http://i.imgur.com/GayhImd.png


Stickymaddness:
More filters are always welcome, I think they could be based out a bit though.

PR #60, opened 2014-12-26
Changed background color of "Advanced Search" checkboxes for visibility

omgftw:
Changed the background color of the checkboxes for "Advanced Search" for visibility, since the foreground property does not appear to work for checkbox color.

Before: http://i.imgur.com/CWN2Pgo.png
After: http://i.imgur.com/V1aGiaF.png

### DIFF
--- a/Procurement/View/StashView.xaml
+++ b/Procurement/View/StashView.xaml
@@ -223,7 +223,7 @@
                             <ScrollViewer VerticalScrollBarVisibility="Visible" CanContentScroll="True" Height="455">
                                 <Grid>
                                     <StackPanel>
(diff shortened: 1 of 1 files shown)


Stickymaddness:
I've merged this via command line due to the conflict, I'm not sure why github hasn't reflected it.

omgftw:
It seems the change on line 228 (current line number in the main branch) didn't get merged, which was the removal of the background attribute for the "None" checkbox. This might be why github is saying there are unmerged changes.

Stickymaddness:
Ah yes, that was probably it. I've removed the attribute.

PR #74, opened 2015-01-27
Improve offline / online status handling when idle or PoE is closed.

Kapps:
Seems to work in my testing, idle time is tracked properly and the process detection works fine for at least the international client. Since the xyz control panel is only supported on the international version, that should be sufficient.

### DIFF
--- a/Procurement/Utility/PoeTradeOnlineHelper.cs
+++ b/Procurement/Utility/PoeTradeOnlineHelper.cs
@@ -2,17 +2,34 @@
 using POEApi.Model;
 using System;
 using System.Collections.Specialized;
+using System.Diagnostics;
 using System.Net;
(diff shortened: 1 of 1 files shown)


Stickymaddness:
I'll take a look at merging this and doing a release for it as soon as I can!

Kapps:
Added a [pull request](https://github.com/Kapps/Procurement/pull/1) from @daviddowdle that depends on this pull and adds support for poe.trade's new offline feature when PoE is closed or the user is AFK.

If you'd prefer them to be separate pull requests, I can look into how to do that (I assume by reverting his pull and having him submit the pull on this repo once the original one is merged?).

Stickymaddness:
Nice work guys, sorry I took so long to get this in!

PR #84, opened 2015-02-21
Show crafted mods in generated images, too

Ornedan:
Some item properties were missing from the saved images template. I copied over crafted mods but left out MTX listing - not having the former is obviously wrong but I'm not certain about the latter.


Stickymaddness:
Nice one, I agree there isn't really a need to display the mtx listing, but the crafted text is essential.

PR #99, opened 2015-04-27
Update HttpTransport.cs

manly:
Enabled gzipped data transfers.


Stickymaddness:
I've tested this a few times, with the change Procurement hangs trying to get characters.

manly:
That truly surprises me! As you can see its 1 line code change. Generally when I get theses issues its because I set https instead of http. Unfortunately, if that is the case, I think I would rather have an https connection without compression over http with compression :|

manly:
Indeed, I just checked on my code (I have a totally independant software) and all my connections to poe-related stuff are https except for 
		private const string DEFAULT_CHARACTERS_OF_ACCOUNT_URL = "http://www.pathofexile.com/character-window/get-characters?accountName={0}";

which is http. But seriously, if you make just that one url http rather than https it should work. I'm not saying its as interesting, but for what I do I need to save every bandwith I can!

PR #115, opened 2015-05-25
Workaround for bug in .NET prior 4.5

Psevdych:
.NET prior 4.5 have a bug in DataContractJsonSerializer. DataContractJsonSerializer.ReadObject method can't deserialize a json string that contains more than 684 non-ANSI characters.

More info: https://social.msdn.microsoft.com/Forums/vstudio/en-US/938156c7-ccb5-4436-833d-d560b9901750/the-2048-bug-of-datacontractjsonserializer-class?forum=wcf

I know, there is no non-ANSI on European servers, but is on Russian.


Psevdych:
Closed due to missed branch.

PR #116, opened 2015-05-26
Workaround for bug in .NET prior 4.5

Psevdych:
.NET prior 4.5 have a bug in DataContractJsonSerializer. DataContractJsonSerializer.ReadObject method can't deserialize a json string that contains more than 684 non-ANSI characters.

More info: https://social.msdn.microsoft.com/Forums/vstudio/en-US/938156c7-ccb5-4436-833d-d560b9901750/the-2048-bug-of-datacontractjsonserializer-class?forum=wcf

I know, there is no non-ANSI on European servers, but is on Russian.

### DIFF
--- a/POEApi.Model/POEModel.cs
+++ b/POEApi.Model/POEModel.cs
@@ -9,6 +9,7 @@ using System.Security;
 using System;
 using POEApi.Infrastructure.Events;
 using System.Runtime.Serialization;
+using System.Xml;
 
 namespace POEApi.Model
 {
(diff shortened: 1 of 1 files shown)


Psevdych:
Added branch for request.

Stickymaddness:
Interesting bug in DataContractJsonSerializer. My concern with this workaround is the increased memory footprint, since Procurement already uses a large amount of memory due to the images, I don't want to degrade memory usage any further.

What I've been wanting to do is replace DataContractJsonSerializer with Json.NET as it is considerably faster.

I'm going to close the pull request due to the memory issue, it is not an issue for western users and you can keep the workaround on your branch for Russian users.

спасибо!

PR #122, opened 2015-07-09
This fixes the current bug where you are not able to update or bump the forums.

bxyoung:
This fixes the current bug where you are not able to update or bump the forums.


Stickymaddness:
Thanks for looking into this!

For future reference please do not update the assembly versions and in particular the  latest-release.txt file, as this is used for the update check - I will quickly have to change the file back to 1.9.2, otherwise people will think there is a new version out!

bxyoung:
Sorry about that!                                                                                                                                                                                                                                                                                                                                        Sent from my BlackBerry 10 smartphone.                                                                                                                                                                                                                From: StickymaddnessSent: Monday, July 13, 2015 5:05 PMTo: Stickymaddness/ProcurementReply To: Stickymaddness/ProcurementCc: bxyoungSubject: Re: [Procurement] This fixes the current bug where you are not able to update or bump the forums. (#122)Thanks for looking into this!

For future reference please do not update the assembly versions and in particular the  latest-release.txt file, as this is used for the update check - I will quickly have to change the file back to 1.9.2, otherwise people will think there is a new version out!

—Reply to this email directly or view it on GitHub.

PR #149, opened 2015-08-03
Prevent SetBuyoutView from Closing By Click

wlfbck:
Currently, when you have the SetBuyoutView open and click (left or right) anywhere in it which is *not* a control, it closes the dialog. Which unfortunately is kinda annoying and disrupts the "workflow". This oneliner prevents that.


Stickymaddness:
Nice change!

PR #155, opened 2015-08-07
Updated authentication for Garena

tappabletaco:


Stickymaddness:
I can't merge this as it will break Procurement for non Garena.

I would suggest either making your own fork for Garena and maintaining it, or working with one of the other developers who maintain garena versions.

PR #246, opened 2015-10-03
Map Level changed to Map Tier

HoLyVieR:
This change in the last patch broke the application. This should fix most of the issue people have now with the application.


HoLyVieR:
For those that can't wait for the fix to get merged, here's a pre-built version available for download : https://www.dropbox.com/s/mf1orpmwker7c2p/ProcurementPatch.zip?dl=0

Stickymaddness:
Thanks for getting a build out for people waiting for a new release.

PR #315, opened 2015-12-14
Updating stash request URIs to work with a GGG update.

aitee:
Fixed an issue with stash requests failing because of a GGG update to the URI format.
It now requires account name as a key/value pair.

This should resolve everybody's problems regarding failure to download the stash on load up.


Stickymaddness:
Beat me to it!

PR #491, opened 2016-02-27
Fix idle detection issue when a user has been logged on for a month.

Kapps:
The original idle detection had a bug where it could think a user was AFK if they were logged into their computer for between (roughly) 1 and 2 months due to tick count looping around. This pull changes them to both be signed so the wraparound is consistent.

### DIFF
--- a/Procurement/Utility/PoeTradeOnlineHelper.cs
+++ b/Procurement/Utility/PoeTradeOnlineHelper.cs
@@ -109,7 +109,7 @@ namespace Procurement.Utility
             };
             GetLastInputInfo(ref inputInfo);
             // Allow for TickCount wrap-around.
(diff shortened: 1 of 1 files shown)


Stickymaddness:
I'm curious how you came across this issue, was it purely from a technical p.o.v or something you actually encountered?

Kapps:
The code for the idle detection feature was taken from PoEWhisperNotifier, and I had a user from there report a bug where they were constantly receiving notifications despite not having PoE minimized / being idle. It's rare to encounter, but at least some people do. Unfortunately a hard one to test, but this is logically correct and a quick search shows it's what other people do as well to work around the issue.

Stickymaddness:
Thanks for the info, I'll merge it as soon as I'm able to work on the next release.

PR #552, opened 2016-03-17
Ascendancy fix

hakkotsu:
Main point of this PR was to fix critical error on Inventory tab due to missing Ascendancy avatars (see first commit), but after noticing tremendous size of main image i just batch compressed all files and saved over 7MB of size which reduced overall size of executable from 9MB to just 2MB.


hakkotsu:
also removed Talisman league mentions from settings

Stickymaddness:
Very nice, thanks for getting the images together. 

I'll merge this as soon as I'm able to do a new release.

thailyn:
I'd recommend squashing the commits which update the images added in this branch (the Ascendancy avatars), so there's only one version in the repo.  Even though git will probably only store the deltas to these files in the .git directory, I'm thinking it'll essentially be like storing both versions of the files, and I wouldn't want to see the repo bloated like that if we could easily avoid it (like it can't be avoided for all the other images modified here, as we'd have to rewrite history).

hakkotsu:
Well, its pointless, since the size of images with every commit is lower that previous, and only overhead we having here is actually 300 KB (see avatars compression). Main image was introduced way before and we cant rewrite history (bad git practice). On my opinion clear history messages are much more important than 300KB (we are talking before git compression here)

hakkotsu:
[Procurement.1.9.8-RC3.zip](https://github.com/Stickymaddness/Procurement/files/178672/Procurement.1.9.8-RC3.zip)
Just in case anyone wants to give it a try

Issue #590, opened 2016-04-21
Any fix coming soon for the currency tab issue ?


Gxtro2001:

Stickymaddness:
Procurement 1.9.9 has been released, which fixes this issue!

https://github.com/Stickymaddness/Procurement/releases/tag/1.9.9

PR #593, opened 2016-04-27
Fixed issue #590

ongobongo:


Stickymaddness:
Thanks for fixing this and putting out an .exe for people to use in the meantime.

PR #594, opened 2016-04-27
Added option to filter maps

ongobongo:


Stickymaddness:
Note that you can also use the regular search for maps. If you type in "map" an autocomplete dropdown will display of all maps currently in your stash.

PR #599, opened 2016-05-13
Added missing Divination Cards

CrossCRS:


Stickymaddness:
Thanks for getting this together, one question - why was "Lantador's Lost Love" removed?

CrossCRS:
It was duplicated, so I removed one of them.

Stickymaddness:
Right you are, well spotted!

PR #600, opened 2016-05-21
Added displaying enchantments to item tooltips

CrossCRS:
I also moved crafted mods up, so they're always under all other mods.

Comparision


Stickymaddness:
This is great, thanks for the contribution!

Issue #650, opened 2016-12-11
Cannot load.


Hiyame:
[09-12-2016 20:47] System.InvalidOperationException: Sekvensen innehåller inga matchande element
   vid System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
   vid POEApi.Model.ProxyMapper.GetOrbType(String name)
[09-12-2016 20:47] ProxyMapper.GetOrbType Failed! ItemType = Remnant of Corruption
[09-12-2016 20:48] Error reading character data for character 'DisBeMe', Exception info: 
[09-12-2016 20:48] System.Exception: Error reading character data for DisBeMe, if you are in offline mode you will need to login and update. If you received this error while logging in, the authenticated session may have expired or bad data has been returned by GGG or a network issue may have occurred - Please try again.
   vid POEApi.Model.POEModel.GetInventory(String characterName, Boolean forceRefresh, String accountName)
   vid Procurement.ViewModel.LoginWindowViewModel.LoadCharacterInventoryItems(Character character, Boolean offline)
   vid Procurement.ViewModel.LoginWindowViewModel.<LoadItems>d__24.MoveNext()
   vid System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   vid System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   vid Procurement.ViewModel.LoginWindowViewModel.<>c__DisplayClass23_0.<Login>b__0()
   vid System.Threading.Tasks.Task.Execute()
[09-12-2016 20:49] Error reading character data for character 'DisBeMe', Exception info: 
[09-12-2016 20:49] System.Exception: Error reading character data for DisBeMe, if you are in offline mode you will need to login and update. If you received this error while logging in, the authenticated session may have expired or bad data has been returned by GGG or a network issue may have occurred - Please try again.
   vid POEApi.Model.POEModel.GetInventory(String characterName, Boolean forceRefresh, String accountName)
   vid Procurement.ViewModel.LoginWindowViewModel.LoadCharacterInventoryItems(Character character, Boolean offline)
   vid Procurement.ViewModel.LoginWindowViewModel.<LoadItems>d__24.MoveNext()
   vid System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   vid System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
[...]

PR #655, opened 2016-12-19
Adding Essences to currencies to reduce exceptions thrown by client.

aydjay:
Solves:
#650 

There are other that could be closed as it's not the only one.

I'm not sure that this is the best thing to do but there is a "Generator" for the essences so I hope people don't have to do it by hand.


Stickymaddness:
Nice generator for the Essences, beats writing it out by hand 👍, however I have a few comments:

- Essences aren't currency, rather create a new type for Essences. ItemFactory can then return a new Essences (checking the typeline for "Essence Of" should suffice). That way the Essence class can be extended to contain other info, like the tier and type, which in turn means a filter can be written for them.

- The Essence generator does not need to be in the Currency class or even in the same solution. Rather create a separate tools solution of which the essence generator can be a part of.

aydjay:
Tell me about the hand writting.....

I wasn't super happy about having the generator in the currency class but I wasn't sure where else to put it.

I'll take your comments onboard and I'll see where I get too with it.

aydjay:
This will add essences as a type of their own as you have suggested.

The essence recipe needs to be added but I don't have sufficient essences in my stash to be able to test this adequately.

I'd like to tackle the recipe portion as a separate stream of work.

I had an exception thrown by the converter when hovering over the essences - i've handled it appropriately. The colour can be changed easily enough.

I added a "Stackable" item so that I didn't have any duplication.

Stickymaddness:
Nice find with the NotImplementedException(), that's probably been sitting around in there since the beginning.

I will probably move the generation out, but for now will just merge. Nice work.

Issue #657, opened 2016-12-27
search problems 


Gxtro2001:
I tried looking for items with the Adds physical damage box clicked. and nothing come up at all . I have the same problem with other item search problems as well. 
I'm I doing something wrong ? 

Thanks for any help .

Stickymaddness:
This is a bug due to GGG changing some of the description text, and will be fixed in the next release.

Issue #627, opened 2016-09-30
Bug in advanced search


Ilcarro1980:
Adds physical, fire, cold, lightning damage and global crit. multiplier doesn't work (they don't show anything)

Stickymaddness:
I'll look into this asap

PR #659, opened 2016-12-28
Fixed adds damage filters. Fixed +x to level of gems filters. Added t…

### LINKED ISSUE #639: No Advanced Search Options for Chaos Damage/Gems
Every element type (fire/cold/lightning) has the following advanced search options:
- Resistance
- Gem Level
- Adds Damage
- Increased Damage

Only chaos resistance can be searched for.  Can the missing chaos damage/gem search options be added?

CrossCRS:
…hree missing chaos-related filters.

This should fix #639 and #627 

By the way, right now ExplicitModBase and StatFilter don't search in CraftedMods, so a lot of filters don't work on items with crafted mods, should it search for mods in crafted mods as well?

### DIFF
--- a/Procurement/Procurement.csproj
+++ b/Procurement/Procurement.csproj
@@ -152,6 +152,7 @@
     <Compile Include="ViewModel\Filters\ForumExport\AccurayFilter.cs" />
     <Compile Include="ViewModel\Filters\ForumExport\CraftedModFilter.cs" />
     <Compile Include="ViewModel\Filters\ForumExport\CurrencyFilter.cs" />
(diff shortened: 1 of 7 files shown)


Stickymaddness:
Nice, thanks for fixing these.

Yeah, ExplicitModBase and StatFilter are from way back when Procurement was originally written, so crafted stuff isn't taken into account - it would be a nice addition though.

aydjay:
Will this fix #657 too? Shall we mark that bug as closed if that's the case?

Stickymaddness:
This will fix it yes, I'll close #657 once the next release is out.

Issue #667, opened 2017-02-17
Legacy League


aydjay:
In an effort to either 

a) Be ready ahead of time for legacy league support
or
b) Be ready very quickly after legacy league starts.

I'd like to make a list of the things that need to be implemented in procurement that I can see from the league release announcement: https://www.pathofexile.com/forum/view-thread/1834597

- [x] Leaguestones - Probably something similar to what was done with the essences
- [x] Relic identification - This would need to check that we can just add in "Relic" in the Rarity enum and handle it appropriately in the ItemToColorBrushConverter. See PR #668 
- [x] Update the settings to search for Legacy League Settings line 84, 85
This league will go live on the 3rd of March.

Stickymaddness:
Leaguestones should be easy to implement, support for this could almost be done in preparation with the assumption that they will act as other league items have in the past

Relic, will hopefully be as simple as updating the enum and brush converter as you saw.

League settings, it's worth noting that this doesn't _have to_ be updated, by default settings if you chose to login to all leagues, new leagues are automatically picked up, periodically updating the settings file is just something I've always done with each update.

aydjay:
Explanation of League Stones by ZiggyD

https://www.youtube.com/watch?v=xp_a226eg_U

PR #673, opened 2017-03-06
Added in Legacy League to the settings and deprecated breach league.

aydjay:
Tested LeagueStones and they work correctly.


aydjay:
This is ready to be merged in - I've tested that you can login and examine the league stones. I don't think we need to do anything else to get them to work, 
The leaguestone recipe is 3 x any league stone will give you a mystery league stone - hardly worth implementing as the outcome is variable.

aydjay:
This closes one of the work items in #667

Issue #680, opened 2017-03-12
Leaguestones don't get updated to forum thread.


Zibokwe:
Leaguestones are not supported, they don't appear in any category when you price them.

aydjay:
You are correct. I have fixed this issue. Thanks for the report.

See:
https://www.pathofexile.com/forum/view-thread/1866871

Fix in #684

Stickymaddness:
This has finally been released as part of Procurement 1.12.0

PR #684, opened 2017-03-20
Proper leaguestone support

aydjay:
Based on the report #680 

I have confirmed that leaguestones do not work in the forum template.

I have updated the forum template and ensured that they get "Visited" correctly.

image

image

image

### DIFF
--- a/POEApi.Model/ProxyMapper.cs
+++ b/POEApi.Model/ProxyMapper.cs
@@ -302,15 +302,24 @@ namespace POEApi.Model
 
         public static ChargeInfo GetCharges(List<JSONProxy.Property> list)
         {
-            var chargeSize = list.Find(p => p.Name == CHARGES);
-            if (chargeSize == null)
-                return new ChargeInfo(0, 0);
+            try
+            {
+                var chargeSize = list.Find(p => p.Name == CHARGES);
+                if (chargeSize == null)
+                    return new ChargeInfo(0, 0);
 
-            var qty = (object[]) chargeSize.Values[0];
+                var qty = (object[]) chargeSize.Values[0];
             
-            var max = (object[]) chargeSize.Values[1];
+                var max = (object[]) chargeSize.Values[1];
 
-            return new ChargeInfo(int.Parse(qty[0].ToString()), int.Parse(max[1].ToString()));
+                return new ChargeInfo(int.Parse(qty[0].ToString()), int.Parse(max[1].ToString()));
+            }
+            catch (Exception ex)
+            {
+                Logger.Log("Error in ProxyMapper.GetCharges: " + ex);
+            }
+
+            return new ChargeInfo(1,1);
         }
     }
 }
--- a/POEApi.Model/ProxyMapper.cs
+++ b/POEApi.Model/ProxyMapper.cs
@@ -9,6 +9,7 @@ namespace POEApi.Model
     internal class ProxyMapper
     {
         internal const string STACKSIZE = "Stack Size";
+        internal const string CHARGES = "Currently has %0 of %1 Charges";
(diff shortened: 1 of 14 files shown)


POEApi.Model/ProxyMapper.cs
```diff
@@ -298,5 +299,18 @@ internal static List<Tab> GetTabs(List<JSONProxy.Tab> tabs)
                 throw;
             }
         }
+
+        public static ChargeInfo GetCharges(List<JSONProxy.Property> list)
```
Stickymaddness: Put a try catch around this method, given how frequently GGG can change things it's safer to catch the exception, log it and return a default value.

POEApi.Model/ProxyMapper.cs
```diff
@@ -298,5 +299,18 @@ internal static List<Tab> GetTabs(List<JSONProxy.Tab> tabs)
...
+        {
+            var chargeSize = list.Find(p => p.Name == CHARGES);
+            if (chargeSize == null)
+                return new ChargeInfo(0, 0);
+
+            var qty = (object[]) chargeSize.Values[0];
```
Stickymaddness: Is the cast to object[] necessary? Is there not a nicer way to do this?

POEApi.Model/ProxyMapper.cs
aydjay: I will put a try catch around the creation of the ChargeInfo object. I will return a new charge(1,1) if it fails.

POEApi.Model/ProxyMapper.cs
aydjay: I will investigate to see if there is a better way to do this.

aydjay: I have checked out the JSon that comes back.

It's a little bit of an odd way of doing it and it's not exactly clear if the first set or the last set are the number of charges or the max charges.

```
	{
		"verified": false,
		"w": 1,
		"h": 1,
		"ilvl": 69,
		"icon": "https:\/\/web.poecdn.com\/image\/Art\/2DItems\/Currency\/Classic\/Prophecy.png?scale=1&w=1&h=1&v=ec3aac908a0be262b312c01c88545e7f3",
		"league": "SSF Legacy",
		"id": "8ab56a1f74390a4454b93fb7e24d42d6243828c7e157d5fc48840040316e9c64",
		"sockets": [],
		"name": "",
		"typeLine": "<<set:MS>><<set:M>><<set:S>>Ample Prophecy Leaguestone of Silver",
		"identified": true,
		"corrupted": true,
		"lockedToCharacter": false,
		"properties": [{
			"name": "Currently has %0 of %1 Charges",
			"values": [["5",
			0],
			["5",
			0]],
			"displayMode": 3
		}],
		"implicitMods": ["Next Area will contain a Silver Coin"],
		"explicitMods": ["Next Area will contain an additional Silver Coin",
		"Monsters with Silver Coins in next Area will drop an additional Silver Coin"],
		"descrText": "Right-click to open your Legacy Panel and place a Leaguestone in an empty slot. It will apply to the next new area you enter.",
		"frameType": 1,
		"x": 1,
		"y": 0,
		"inventoryId": "Stash8",
		"socketedItems": []
	}
```

As you can see the properties are a pairing of string and int. So it's an array with mixed types. I can see the concern with this loose object[] cast but I can't think of a better way. I'm open to suggestions. Atleast this method is not wrapped with a try catch and will return a "Sensible" object.

Stickymaddness: That is rather odd, but then that's the GGG api for you ;)

Thanks for looking into the json, looks like we'll have to stick with this for now!

PR #706, opened 2017-08-01
Removed Old Json Data Contract Serialiser in favour of using Json.Net

aydjay:
After a scant 542 days, I have finally gotten round to using Json.Net throughout the entire application

It feels to me to be loading the application in offline mode more quickly but I suspect the increase in speed is merely a placebo.


Stickymaddness:
This is a loooooooooooooooong overdue change, I wouldn't be surprised if loading offline (and online) is significantly faster as NewtonSoft is significantly faster than the old data contract serializer.

PR #736, opened 2017-09-08
Fix: Morning Stars aren't recognized as maces.

zorbathut:


aydjay:
Looks good to me!

Could you add a supporting unit test too please?

Issue #737, opened 2017-09-10
Unable to cast object of type when downloading inventory


lythix:
[DebugInfo.zip](https://github.com/Stickymaddness/Procurement/files/1290305/DebugInfo.zip)

aydjay:
In the application installation folder there will be a folder with your profile name on it. Inside there will be a load of bin files. Can you please zip up the bin files for me and I will try and replicate and resolve this issue.

lythix:
Cool, miss it. There's nothing in the folder that can reveal my account right? Especially passwords
[procurement.zip](https://github.com/Stickymaddness/Procurement/files/1290701/procurement.zip)

aydjay:
#738 should fix you @lythix.

Please pull it and try it both online and offline.

lythix:
Hmm, no go. Pulled your branch from your repo and tried it. Will debug it later and see if I can see what's going on but it still had some issues when in online mode.

aydjay:
My hunch is it has something to do with characters Mortrend and Messycrime scene.

If you could look at the GetCharacters when in online mode you may see that there are extra properties which are important but we are ommiting.

lythix:
It doesn't like charges...


[cid:af6f43db-cd20-4c7d-963f-fcae064e9a49]


It hit that much earlier than first try, so it probably just ignores this one...

________________________________

aydjay:
Have a look at proxymapper.cs in: https://github.com/Stickymaddness/Procurement/pull/738/commits/c8baf141e7af87c7218214fe982cbb5b7ec8b573

Bear in mind that this isn't the end off the world as it will log an exception and give you a stack size of 1/1

aydjay:
Can you try again with the latest version please

lythix:
Will do. Sorry have been crazy busy. Will report back how it goes. Thanks!

Stickymaddness:
Please test with [Procurement 1.20.0](https://github.com/Stickymaddness/Procurement/releases/tag/1.20.0), and create a new ticket if this is still an issue.

PR #738, opened 2017-09-11
Chached transport offline understanding

aydjay:
While investigating #737 

I discovered something odd - when loading in offline mode - it sometimes tries to get the online data!

Not cool.

The issue from #737 was 2 fold. There is a character in the character.bin (Mortrend) that shows in SSF Standard. But there aren't any corresponding tabs for SSF Standard. This causes it to keel over. This is now fixed

Secondly - the character Messycrimescene didn't have any inventory data - this cause it to try and go online to find it. But in offline mode it keels over.

Additionally it looks like the "ChargeInfo" from the server is a pickle. I've improved it once again in the hope that it's fixed properly.

Let me know what you think


lythix:
Will have a look and test. Didn't even know there was an offline mode.

POEApi.Model/POEModel.cs
```diff
@@ -188,7 +195,8 @@ public Stash GetStash(string league, string accountName)
 
                 Stash stash = GetStash(0, league, accountName);
 
-                if (stash.Tabs[0].Hidden)
+                var firstTab = stash.Tabs.FirstOrDefault();
+                if (firstTab != null && firstTab.Hidden)
```
Stickymaddness: This could be simplified as:

```c# 
firstTab?.Hidden
````

File: POEApi.Model/POEModel.cs [hunk 2]
```diff
@@ -239,7 +247,8 @@ private Stash getAllTabs(string league, string accountName, Stash stash)
                 }
             }
 
-            if (stash.Tabs[0].Hidden)
+            var firstTab = stash.Tabs.FirstOrDefault();
+            if (firstTab != null && firstTab.Hidden)
```
Stickymaddness: Same here

PR #742, opened 2017-09-13
Fix: Trisula not recognized as a knife, missing unit test for Morning Star.

zorbathut:


POEApi.Model/Data.xml
```diff
@@ -634,6 +635,7 @@
       <Item name="Sambar Sceptre"/>
       <Item name="Behemoth Mace"/>
       <Item name="Coronal Maul"/>
+      <Item name="Morning Star"/>
```
aydjay: I can see the morning star getting added here but I don't see a supporting unit test for either of them :/

zorbathut:
Dang it, thought that was what was needed for the unit test. Which file should I look at? I had trouble finding it.

aydjay:
Have a look at the IFilterTest - that should give you an idea of how the unit tests have been done.

zorbathut:
Hmm. Looks like the best solution is to just take a dump of a stash with the items; unfortunately, I no longer have any trisulas or morning stars. Want me to fake some items in a stash, or just wait until I have 'em again?

aydjay:
I'm not super keen on faking them myself - just because a real world test is best. If you are able to create an appropriate unit test then great - otherwise it's no big deal. There isn't a single "Runner" that has a supporting unit test as it stands. I'm just trying my best to support any new functionality with an appropriate Unit test. :)

Stickymaddness:
I'm going to merge this in the interest of getting the feature released sooner, but a supporting test in a follow up PR would be _greatly_ appreciated. 

One of my failings with Procurement was not writing tests from the get go.

zorbathut:
I'll see what I can do, as soon as I find another trisula :)

PR #744, opened 2017-09-15
Added Offering of the goddess support

### LINKED ISSUE #739: No Offering to the Goddess
Hey,i can see the "Offering to the Goddess" on Procurement stash tab but can't see it on the forum thread.Bumped it,updated it and restarted the Procurement several times.

aydjay:
This depends on #740 as it uses the same sample data in the unit test.

This will close #739


Stickymaddness:
You are killing it with these updates aydjay! 

I'll deploy a release now.

PR #757, opened 2017-10-02
better handling of divination cards

MarcLandis:
The cards are now added to a normal stash using extra code. I will work on adding a real card view like the web site has. 
The item hover popup has been enhanced to show the cards like the web.


POEApi.Model/POEModel.cs
```diff
@@ -300,6 +300,11 @@ private void getImageWithEvents(Item item)
             try
             {
                 getImageWithEvents(GetItemName(item), item.IconURL);
+
+                if (item.ItemType == ItemType.DivinationCard)
```
aydjay: it will do it twice for divcards which is inefficient

if (Divcard)
   get()
else
   get()

POEApi.Model/POEModel.cs
```diff
@@ -300,6 +300,11 @@ private void getImageWithEvents(Item item)
             try
             {
                 getImageWithEvents(GetItemName(item), item.IconURL);
+
+                if (item.ItemType == ItemType.DivinationCard)
+                {
+                    getImageWithEvents(GetItemName(item), ((DivinationCard)item).ImageURL);
```
aydjay: Might throw if it isn't actually a div card

POEApi.Model/ProxyMapper.cs
```diff
@@ -200,6 +200,192 @@ internal class ProxyMapper
...
+            {"Three Faces in the Dark", DivinationCardType.ThreeFacesInTheDark},
+            {"Thunderous Skies", DivinationCardType.ThunderousSkies},
+            {"Time-Lost Relic", DivinationCardType.TimeLostRelic},
+            {"Tranquillity", DivinationCardType.Tranquillity},
+            {"Treasure Hunter", DivinationCardType.TreasureHunter},
+            {"Turn the Other Cheek", DivinationCardType.TurnTheOtherCheek},
+            {"Vinia's Token", DivinationCardType.ViniasToken},
+            {"Volatile Power", DivinationCardType.VolatilePower},
+            {"Wealth and Power", DivinationCardType.WealthAndPower}
+        };
```
aydjay: Is there a way we can omit the need to maintain long lists of all possible divination cards? I never liked the way it was done.

Procurement/View/Converters/DivinationCardExplicitModsToFormattedRunConverter.cs
```diff
@@ -0,0 +1,84 @@
...
+            var runs = new List<Run>();
+
+            var count = 0;
+            
+            foreach (Match part in Regex.Matches(mod, "<(?<type>.*?)>\\{(?<text>.*?)\\}"))
+            {
+                switch (part.Groups["type"].Value)
+                {
+                    case "default":
+                        runs.Add(new Run(part.Groups["text"].Value + (count > 0 ? " " : "")) { Foreground = (SolidColorBrush)new BrushConverter().ConvertFrom("#7f7f7f") });
```
aydjay: Maybe consolidate and standardise the Hexcodes for each item.

Have a look at "ItemToColorBrushConvertor.cs"

Consolidating the "Colors" in the App.Xaml as StaticResources would be the way to go.

POEApi.Model/Item.cs
```diff
@@ -53,6 +54,7 @@ public abstract class Item : ICloneable
         public string TradeInventoryId { get; set; }
         public string Character { get; set; }
         public int ItemLevel { get; set; }
+        public string ArtFilename { get; set; }
```
aydjay: ArtFilename is a consideration only of divination cards - this should be in DivinationCard.cs instead as it will be null for everything else.

POEApi.Model/POEModel.cs
MarcLandis: the first call will get the icon for the stash and the second the actual image for the popup

Stickymaddness:
@MarcLandis Do you still intend to update this PR?

KaomsHeart:
I am currently short on time, but try to get this done.

PR #762, opened 2017-10-24
How to get SessionID?

bule2015:
Added the image to jump to a page with instructions on how get the SessionID.

[https://gyazo.com/50e9d2069d67bce5247ddc240a92138e](https://gyazo.com/50e9d2069d67bce5247ddc240a92138e)


Stickymaddness:
I like the idea of this, but the wiki article is _ancient_, I don't even think the cookie name it refers to is correct. In order to merge something like this there would need to be an up to date article on the github wiki that it links to.

bule2015:
I expected that someone would correct wiki. 
Because I'm because there are no English abilities fatally.
But I'll try my best.

bule2015:
How about with this?
https://github.com/bule2015/Procurement/wiki

aydjay:
Looks good apart from the steam section - it reads a little bit like you can login to procurement with steam, which isn't the case.

I would write:

・Can I log in to pathofexile.com using my steam account?

　　Yes, Please login using "Sign in through STEAM". The POESESSID retrieval will be the same.

bule2015:
It was rewritten as follows.

・Can Steam account also acquire SessionID?
　　Yes, When logging in by a official site, SessionID can be acquired.

Stickymaddness:
I've copied your wiki page to https://github.com/Stickymaddness/Procurement/wiki/SessionID, if you link to it in your PR I can merge it.

PR #763, opened 2017-10-29
A push on a key will display all the sockets

bule2015:
When an ALT key is pressed, all sockets are indicated.
When an ALT key is separated, indication goes off.

[https://gyazo.com/4ef250b86203f82130feaa7e39a6f95e](https://gyazo.com/4ef250b86203f82130feaa7e39a6f95e)


aydjay:
Great change!

Issue #494, opened 2016-03-01
Add a LICENSE file


thailyn:
In #194, you stated the license for Procurement is Artistic License 2.0, and it's documented in the about screen.  Can we have a LICENSE or LICENSE.txt file in the repo stating this and making it explicit?  I would need such a file present before contributing any more code (it's been a while...).

In terms of what that license should be, I trust you understand licenses well.  I, however, do not, especially when trying to get them to interoperate.  My interpretation of the GPLv2 (which is probably wrong) brings a few complications to mind:
- According to this [list of licenses](https://www.gnu.org/licenses/license-list.html#GNUGPL), the Artistic License 2.0 [is compatible](https://www.gnu.org/licenses/license-list.html#ArtisticLicense2) with the GNU GPL licenses.  Thus, all of the code except for RequestThrottle can be licensed under the Artistic License 2.0.  However, according to [this FAQ question](https://www.gnu.org/licenses/gpl-faq.html#WhatDoesCompatMean), the combined work (Procurement as a whole) must be distributed under the GPLv2.  I could be wrong, but that's how I'm reading it.
- You modified the RequestThrottle class, even if simply to put it in a namespace to fit in with the rest of Procurement, but section 2a of the [GPLv2](https://tldrlegal.com/license/artistic-license-2.0-%28artistic%29#fulltext) says, "you must cause the modified files to carry prominent notices stating that you changed the files and the date of any change," and I don't see any such notice in the file.

My apologies for perhaps splitting hairs on all this, but I just want to make sure there isn't any possibility of license complications with these multiple licenses intermingling.

Stickymaddness:
Hi Thailyn,

I actually do not understand licenses well myself at all. Case in point, I did not realise the namespace addition in RequestThrottle required a notice explaining the change, nor did I consider that multiple licenses might be a problem.

Prior to Procurement I had no experience with creating an open source project, and certainly did not expect it to become as big as it did or still be going 3 years down the road. I just wanted to share Procurement and picked a license that google code supported that seemed to me to be the most straight forward.

Point taken on what you have brought up above though.

I'm happy to add a LICENSE.txt with the Artistic License 2.0, and will need to look into the potential issues with GPLv2/RequestThrottle that you have brought up.

thailyn:
Hi, Stickymaddness.

I'm glad I'm not the only one whose brain hurts just thinking about licenses. :)

[...]

PR #811, opened 2018-02-21
Update task throttle

thailyn:
As detailed in #494, this replaces the RequestThrottle class with a simpler version, and then adds license files and other small changes in an effort to make the licenses of the code and binaries in this repository clear.


aydjay:
I can't comment about #494 as I'm not well enough versed in legalese.

With regards to adding licence files for WPFToolkit from codeplex. Again - I'm not well versed on the legalese. We shouldn't however be including a Lib folder with a DLL. We should manage this dependency using NuGet as WPFToolkit is available through that.

This may/may not negate the need for a licence file all together (Again - I don't know anything about licences).

The dll isn't included in our repo through the leverage of NuGet. Please correct me if i'm wrong.

Procurement/View/AboutView.xaml
```diff
@@ -65,9 +65,7 @@
                         <TextBlock />
                         <TextBlock Foreground="#FFAB9066" TextWrapping="Wrap" Text="Path Of Exile, the Path Of Exile logo and all related artwork used in this application are Copyright © Grinding Gear Games. For further information please visit:                     http://www.pathofexile.com" FontSize="11" />
```
aydjay: Introduce a style so that repeated properties aren't duplicated.

Stickymaddness:
@aydjay You are correct about the wpf toolkit, hell I'd **_completely_** forgotten about that. It should definitely be moved to nuget.

PR #813, opened 2018-03-01
Fix rare set recipe

thailyn:
I think we can further clean up this recipe class, but this is a first step for the moment.  I'm mostly concerned at the moment that it is correct.

This also sorts items by item level, so lower-level (and presumably worse) items are picked before higher-level (and presumably better) ones, among those in the item level range allowed for the recipe.


aydjay:
Looks good - any chance of some unit tests?

thailyn:
Last time I contributed the project just didn't have any unit tests.  I'll take a look this weekend.

thailyn:
Sorry for the delay on the unit tests.  I started taking a look; I'll send out a new PR with some tests soon.

PR #814, opened 2018-03-02
Add new item types through 3.1

thailyn:
Various base item types have been added to the game but haven't been added to Procurement.  I've included all the ones I noticed are missing (and there will be a few more after 3.2, which I'll take a look at after it releases).  In particular, this adds Jewel and Abyss Jewel base item types to the data xml file, so items of these types have a non-null base type.

I've also fixed all the misspellings of "Abyss Jewel".  It is confusing remember which terms introduced in the Abyss League use "Abyss" versus "Abyssal".


aydjay:
Looks good to me :)

PR #815, opened 2018-03-02
Update same name recipe.

thailyn:
This pull request simplifies and improves the same name recipe, and allows Abyss Jewels to be included in the results, as detailed in the commit messages.

The way the rarity of gear and abyss jewels is checked separately here is clunky, but I have some ideas I can look in to later to clean up (e.g., reorganizing some classes so they inherit Rarity from a common parent).


aydjay:
Looks good to me - Unit tests? ;-)

PR #816, opened 2018-03-02
Include number of complete recipes in description.

thailyn:
Update the description for each section in the list of recipe results to show both the number of completed (100%) recipes, in addition to the normal number of total recipes.  This makes it easier to know at a glance if a section has any recipes that can be acted upon, instead of only knowing if there are recipes that are either close or complete.


aydjay:
Looks good to me +1

PR #818, opened 2018-03-12
Reorganize tests and add SameNameRecipe tests.

thailyn:
Several objectives in this PR:
* Move the existing test project into a Tests subdirectory (we should have a whole bunch of test projects, so let's tuck them away from normal view).
* Moved the JSON-builder helper classes into a TestsHelpers project.
* Added a project Procurement.Tests with tests for classes in the Procurement project.
* Added a RecipesTest class with tests (as an initial version) for the SameNameRecipe class.  These cover the changes in #815.

As I referenced in one of the commit messages, there are a lot of classes or methods which are private or internal.  I don't see a reason for them to not be public, so I have been changing them to public, as needed, to write tests.


aydjay:
Nice work!

PR #822, opened 2018-03-15
Sort compatible types to use longer base types.

thailyn:
There are base types in the same gear type where one base type is a
non-trivial substring of another base type.  For example, there are
"Tricone", "Noble Tricorne", and "Sinner Tricorne" base types.  Since we
determine and item's base type by checking if each base type is a
substring of that item's type line (since the type line can include
affixes), we need to make sure we do not use a shorter base type when a
longer one is more correct.  The simple fix used here, after building
the list of compatible types for a gear type, sort that list by string
length, so the longer strings are checked first.

### DIFF
--- a/POEApi.Model/GearType/GearTypeRunner.cs
+++ b/POEApi.Model/GearType/GearTypeRunner.cs
@@ -25,7 +25,7 @@ namespace POEApi.Model
             : base(gearType)
         {
             this.generalTypes = new List<string>();
-            this.compatibleTypes = compatibleTypes.ToList();
+            this.compatibleTypes = compatibleTypes.OrderByDescending(s => s.Length).ToList();
             this.incompatibleTypes = new List<string>();
         }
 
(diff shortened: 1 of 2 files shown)


aydjay:
This is resolving the side effect rather than the base issue imo.

https://github.com/Stickymaddness/Procurement/blob/669db75cfbe28dce98ae3c5e43d1cd91b1ea6c0c/POEApi.Model/GearType/GearTypeRunner.cs#L32

This method uses contains - which causes the above behaviour. Might be better to use equals and StringComparison.CurrentCultureIgnoreCase

thailyn:
Using a test for string equivalence won't work here because the type line for magic items includes the affixes (I should add a test with that).

I agree it's not an ideal solution, but it's a simple change that fixes the bug, and it's generic (it won't need to be updated as new items are released, unlike, for example, the lists of incompatible types). I haven't come up with a more elegant solution yet.

The whole process of checking substrings to determine base types isn't ideal, either, but for the most part, it works. I'd like to take a closer look at all the data we're getting from the API to see if there's a better solution (particularly now that we can get the full affix breakdown in the client), but I'm still working on low-hanging (and user-visible) changes.

Stickymaddness:
It might be worth looking into the API, there's probably some new fields that Procurement is missing that might help in this regard.

Merging this as I agree that this is a quick win.

PR #835, opened 2018-05-20
Handle necromancy nets

### LINKED ISSUE #832: League Selection
I am using this software for about a week without any issues, but now I set some rings for sale, but it started crashing afterwards, with the error message shown below. Now every time I try to open the program, it works, but it crashes constantly when selecting the league (BRE001)

thailyn:
Necromancy Nets do not have a net tier, so right now Procurement fails to create a `Net` item for Necromancy Nets.  It does successfully create an `UnknownItem`, but, since this item type has basically no properties set, this "fake" item does not show up in the tab/inventory UI (since `InventoryId` is null) and can cause exceptions or crashes elsewhere (e.g., in the various item visitors which assume the `TypeLine` is not null).

So, first, this PR properly creates Necromancy Nets, as it does not balk at a missing `NetTier` property.  And, second, when items do fail to be created, try to populate the resulting `UnknownItem` with more information (by passing the underlying `JSONProxy.Item` object to the parent `Item` constructor), which will allow that item to show up in the UI and prevent other errors.

Note: I expect this PR will have merge conflicts with #834.  I'll rebase one onto the other after one is merged.

This PR fixes #832.

### DIFF
--- a/POEApi.Model/ItemFactory.cs
+++ b/POEApi.Model/ItemFactory.cs
@@ -55,7 +55,23 @@ namespace POEApi.Model
                 var errorMessage = "ItemFactory unable to instantiate type : " + item.TypeLine;
                 Logger.Log(errorMessage);
 
-                return new UnknownItem();
+                try
+                {
+                    // Try to fall back and create an unknownItem based off of the provided item object.  This will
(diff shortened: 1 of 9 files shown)


thailyn:
Resolved merge conflicts.

PR #837, opened 2018-05-21
Added visitor and filter for UnknownItem.

thailyn:
This is mostly as a debugging tool.  If any items show up in this filter, then there's an outstanding bug with Procurement that needs to be fixed.

Note that, without 2c3d7f4 in #835, `UnknownItem` objects will always have a null `InventoryId`, and therefore never show up in the stash/inventory UI, and will not be found by this filter.  Compared to master, with only this PR and #835 merged, this filter will show items such as Pantheon Souls (but not after #834 is merged).


aydjay:
Tempted to recommend an #IF DEBUG compiler directive here but this may well come in handy in the field for diagnosis.

👍

PR #838, opened 2018-05-21
Revise flavour text and ItemHover handling

thailyn:
The `FlavourText` property was only being given to some subclasses of `Item`, but that missed some item types, such as Pantheon Souls.  This PR promotes `FlavourText` to be a property of `Item`, and it is displayed in the item tooltip whenever it is present.

This PR also reworks how the sections of the item tooltip are assembled.  It is more correct now in terms of being more inclusive.  There are still some known issues, including having an extra divider at the bottom of the tooltip, when the item does not have description text.  Since it more correctly displays the actual text, I think this is an improvement for now.  I have some ideas for how to further improve the item tooltip and handle these known issues.  I don't like working with GUIs, though....

### DIFF
--- a/Procurement/View/VisibilityConverter.cs
+++ b/Procurement/View/VisibilityConverter.cs
@@ -30,7 +30,7 @@ namespace Procurement.View
 
                     if (enumerable != null)
                     {
-                        result = enumerable.Cast<object>().Any() ? Visibility.Hidden : Visibility.Visible;
+                        result = enumerable.Cast<object>().Any() ? Visibility.Visible : Visibility.Hidden;
                     }
                     else  if (value is int)
                     {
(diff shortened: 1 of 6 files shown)


Procurement/View/VisibilityConverter.cs
```diff
@@ -30,7 +30,7 @@ public object Convert(object value, Type targetType, object parameter, System.Gl
 
                     if (enumerable != null)
                     {
-                        result = enumerable.Cast<object>().Any() ? Visibility.Hidden : Visibility.Visible;
+                        result = enumerable.Cast<object>().Any() ? Visibility.Visible : Visibility.Hidden;
```
aydjay: This is inverting the logic that is already in place - is this intended? May have repercussions where it is already in use. 

I tend to favour Collapsed compared to Hidden as I don't want the space reserved for the element when it's not visible.

PR #839, opened 2018-05-21
Add a DebuggerDisplay attribute to the Tab class.

thailyn:
It helps with debugging.  ¯\\\_(ツ)\_/¯

### DIFF
--- a/POEApi.Model/Tab.cs
+++ b/POEApi.Model/Tab.cs
@@ -14,6 +14,7 @@ namespace POEApi.Model
         Unknown,
     }
 
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
     public class Tab
     {
         public bool IsFakeTab { get; set; }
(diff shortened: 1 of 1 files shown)


aydjay:
👎 

You never know when the "ToString()" is being called, either implicitly or explicitly. Introducing the method down the road can change previously known and working behaviour.

Considering this is being used explicitly for debugging use the [DebuggerDisplay] attribute on the class instead.

See:
https://msdn.microsoft.com/en-us/library/system.diagnostics.debuggerdisplayattribute.aspx

Best practices:
https://blogs.msdn.microsoft.com/jaredpar/2011/03/18/debuggerdisplay-attribute-best-practices/

thailyn:
Thanks for the links.  I get what you mean about unintended consequences, but I can't think it would be worse than having `{POEApi.Model.Tab}` if we unintentionally printed the `ToString()` output somewhere.

I've switched it to use `DebuggerDisplay`.

PR #840, opened 2018-05-21
Added support to refresh only used tabs.

thailyn:
All tabs until the final non-empty tab is refreshed.

I particularly enjoyed this optimization when I was only using fewer than 42 tabs, but owned more tabs.  In this case, refreshing all tabs would get throttled, but refreshing only used tabs would not and was much quicker.  Now I just hoard way too many items.

### DIFF
--- a/POEApi.Model/Stash.cs
+++ b/POEApi.Model/Stash.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace POEApi.Model
 {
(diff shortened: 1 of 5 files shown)


POEApi.Model/Stash.cs
```diff
@@ -64,6 +65,34 @@ public void RefreshAll(POEModel currentModel, string currentLeague, string accou
             }
         }
 
+        public void RefreshUsedTabs(POEModel currentModel, string currentLeague, string accountName)
+        {
+            var usedTabIds = items.GroupBy(i => i.InventoryId)
+                .Select(group => group.First())
+                .Select(item => item.InventoryId)
+                .Where(id => !string.IsNullOrWhiteSpace(id))  // An UnknownItem could have an invalid InventoryId.
+                .Select(id => Int32.Parse(Regex.Match(id, @"\d+").Value) - 1);
```
aydjay: Difficult to read and digest what this linq is doing. Regex.Match(bla) may throw a null ref for the .Value.

thailyn: The real annoyance with LINQ expressions is that they are fantastic for doing something complicated in a really compact way, but they are both obtuse and do not lend themselves to error checking.  They look great on the outside, but they're not good for you.  The candy of programming patterns.

aydjay: Agreed.

PR #841, opened 2018-05-23
Add missing items

thailyn:
I went through all of the item types in Data.xml and added in any missing items, and sorted a bunch of categories, so they are easier to maintain going forward.  Sort order is first by category (e.g., one-handed vs two-handed), then by level requirement, then alphabetically.

As mentioned in the commit message for 79b1fd8 I have some duplication in the list of maps, when different versions of maps (e.g., Atlas of Worlds vs. War for the Atlas) have the same map name, but I thought that was preferable instead of including more sparse lists that are harder to examine for missing base types.  I removed the unique maps from the list, since, like with unique items for other gear types, the unique maps are a variation of a base type.  I did, however, leave in Atziri's and the Pale Council's maps; as I understand these are map items with their own unique base types.

I might be missing one divination card, as the wiki says there are 201, but there are only 200 in its table of divination cards.  There might be a technical issue and the table stops at 200 items, however, making it not show up on the wiki.  Divination cards that I added in this PR (since it is hard to tell after sorting the list):

    Atziri's Arsenal
    Blessing of God
    Forbidden Power
    Mawr Blaidd
    Might is Right
    Rebirth
    Struck by Lightning
    The Breach
    The Deceiver
    The Dreamer
    The Insatiable
    The Iron Bard
    The Oath
    The Obscured
    The Opulent
    The Puzzle
    The Scavenger
    The Stormcaller
    The Valkyrie
    The Wolf
    The World Eater

I also added the missing FishingRod gear type and related filter/runner.  You know, the important stuff.

### DIFF
--- a/POEApi.Model/Data.xml
+++ b/POEApi.Model/Data.xml
@@ -1701,5 +1701,9 @@
       <Item name="Murderous Eye Jewel"/>
       <Item name="Searching Eye Jewel"/>
     </GearBaseType>
+
+    <GearBaseType name="FishingRod">
+      <Item name="Fishing Rod"/>
+    </GearBaseType>
   </GearBaseTypes>
 </Data>
--- a/POEApi.Model/Data.xml
+++ b/POEApi.Model/Data.xml
@@ -2,31 +2,65 @@
 <Data xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="Data.xsd">
   <GearBaseTypes>
     <GearBaseType name="Ring">
-      <Item name="Iron Ring"/>
+      <Item name="Breach Ring"/>
       <Item name="Coral Ring"/>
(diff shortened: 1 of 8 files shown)


aydjay:
Just referencing this PR: #757 

The way we maintain a static list of "Known Cards" is very poor. The json representation of a divination card has enough information for us to do  away with it.

aydjay:
Sadly it still leveraged the concept of having static data but it had all the UI changes needed to have a div card stash. Worked well IIRC. It's worth resurrecting for sure @MarcLandis. I'm not sure on the full ramifications of working a different solution to having a known list of cards would have but it would certainly reduce some of the development overhead.

Stickymaddness:
It'd be great to have that functionality added to Procurement, I don't think you'd be stepping on anyone's toes at this point as the PR does appear to be abandoned.

PR #844, opened 2018-05-25
Fix login hash regex and add login parameters.

### LINKED ISSUE #843: Unable to log in -- hash format changed
Starting today I can not log in using email and password.  Perhaps this is related to the [GDPR changes](https://www.pathofexile.com/forum/view-thread/2148460/page/1) that just went into effect.  I'm able to log in to the website in a browser without a problem, and can launch the game.  I did, however, need to accept the ToS the first time I logged in to the website, so regardless of any other issue, users might need to accept the ToS manually before they can log in with Procurement.

I first noticed that the regex to [find the login hash](https://github.com/Stickymaddness/Procurement/blob/0a4eb76ca02e5ffaae3d9a71b4e0bbfcbc99932f/POEApi.Transport/HttpTransport.cs#L29) is not correct.  I never looked at this before, so I don't know if this is a new issue.  The hash now seems to have two parts, which are separated by a hyphen.

Current regex:

    private const string hashRegEx = "name=\\\"hash\\\" value=\\\"(?<hash>[a-zA-Z0-9]{1,})\\\"";

Correct regex:

    private const string hashRegEx = "name=\\\"hash\\\" value=\\\"(?<hash>[a-zA-Z0-9-]{1,})\\\"";

Even after changing the regex the [login response](https://github.com/Stickymaddness/Procurement/blob/0a4eb76ca02e5ffaae3d9a71b4e0bbfcbc99932f/POEApi.Transport/HttpTransport.cs#L110) is 200 OK instead of 302 Found.

Information from the `HttpWebResponse` object:

    {System.Net.HttpWebResponse}
        CharacterSet: "UTF-8"
        ContentEncoding: ""
        ContentLength: -1
        ContentType: "text/html; charset=UTF-8"
        Cookies: {System.Net.CookieCollection}
        Headers: {Transfer-Encoding: chunked
    Connection: keep-alive
    Pragma: no-cache
    X-Frame-Options: SAMEORIGIN
    Cache-Control: no-store, no-cache, must-revalidate
    Content-Type: text/html; charset=UTF-8
    Date: Fri, 25 May 2018 01:21:01 GMT
    Expires: Thu, 19 Nov 1981 08:52:00 GMT
    Server: nginx/1.12.2
    
    }
        IsFromCache: false
        IsMutuallyAuthenticated: false
        LastModified: {5/24/2018 9:22:42 PM}
        Method: "POST"
        ProtocolVersion: {1.1}
        ResponseUri: {https://www.pathofexile.com/login}
        Server: "nginx/1.12.2"
        StatusCode: OK
        StatusDescription: "OK"
        SupportsHeaders: true

I never looked at this code or the login process before, so I don't know if anything's changed, and nothing is jumping out at me that signals what could be going wrong.

thailyn:
This PR fixes the regex that looks for the hidden `hash` form field, since it appears to now be a longer string with a hyphen separating two alphanumeric (hexadecimal?) strings.  It also adds two more parameters to the login request.  Adding and fixing these request parameters is sufficient to allow me to log in again, but I'm not certain they are all necessary.

Using `&remember_me=1` appears to not work, and I am not sure if this means the Procurement "user" will be logged out after some time and stop working.  However, I think `&remember_me=0` is the default value, so hopefully everything works the same as before.

So, this PR appears to fix #843.  I haven't tested this for any length of time, but I wanted to at least create this PR so it is ready, particularly if other users start having problems logging in.

### DIFF
--- a/POEApi.Transport/HttpTransport.cs
+++ b/POEApi.Transport/HttpTransport.cs
@@ -26,7 +26,7 @@ namespace POEApi.Transport
         private const string characterURL = @"https://www.pathofexile.com/character-window/get-characters";
         private const string stashURL = @"https://www.pathofexile.com/character-window/get-stash-items?league={0}&tabs=1&tabIndex={1}&accountName={2}";
         private const string inventoryURL = @"http://www.pathofexile.com/character-window/get-items?character={0}&accountName={1}";
-        private const string hashRegEx = "name=\\\"hash\\\" value=\\\"(?<hash>[a-zA-Z0-9]{1,})\\\"";
+        private const string hashRegEx = "name=\\\"hash\\\" value=\\\"(?<hash>[a-zA-Z0-9-]{1,})\\\"";
 
         private const string updateThreadHashEx = "name=\\\"forum_thread\\\" value=\\\"(?<hash>[a-zA-Z0-9]{1,})\\\"";
         private const string bumpThreadHashEx = "name=\\\"forum_post\\\" value=\\\"(?<hash>[a-zA-Z0-9]{1,})\\\"";
@@ -93,7 +93,9 @@ namespace POEApi.Transport
             StringBuilder data = new StringBuilder();
             data.Append("login_email=" + Uri.EscapeDataString(email));
             data.Append("&login_password=" + Uri.EscapeDataString(password.UnWrap()));
+            data.Append("&remember_me=0");
             data.Append("&hash=" + hashValue);
+            data.Append("&login=Login");
 
             byte[] byteData = UTF8Encoding.UTF8.GetBytes(data.ToString());
 


aydjay:
👍

thailyn:
We have a confirmation this is happening for other people in #846, so it's probably at least preventing everyone who uses email/password from logging in.  I also didn't encounter any issues/regressions from this PR while I was playing a bit this evening.  @Stickymaddness, I hate suggesting what you could do, but would you be able to create something like a 1.22.0-RC release ASAP with this PR?  I'm planning on working on a bunch of stuff this weekend, so maybe a full release during next week, before the next league starts?

Stickymaddness:
I'll get through the PRs now, and do a release. We can always do another release during the week if you have a bunch more PRs

PR #848, opened 2018-05-26
Update drop only gems

thailyn:
The list of drop-only gems was outdated and some gems were misnamed.  I've updated the list, and moved it into the Settings.xml file, so it can be more easily maintained in the future.


Stickymaddness:
Might be better suited to put these into data.xml in the future

thailyn:
Yeah, I was considering that.  But I have a feeling the "popular" gems is horribly out of date, so these lists need to get taken a look at again.

PR #849, opened 2018-05-26
Option to skip refresh on login

thailyn:
This PR adds an option on the login window to enable/disable forcing refreshing an account's stash tabs.  If the option is disabled, only missing tabs are downloaded while logging in.  This can help particularly in two cases.  First, if the user wants to close and reopen Procurement and not have to reload all tabs, which could be up to date anyway.  Second, if the user has Procurement keep track of multiple leagues, but the user is only going to reference one of those during a play session, he can have Procurement skip refreshing the tabs from all of those leagues, and manually refresh the tabs for the relevant league, after the login process is completed.

However, since the refresh process that can happen during login deletes and rebuilds the entire folder of `bin` files, if the user skips refreshing tabs at login, Procurement will never pick up changes to tab names/colors and tab ordering. (This is more a bug in how tabs are refreshed after logging in.)

I've also added in this PR some more error checking in places, and cleaned up how we determine we need to save the user's settings after modifying the login window's fields.

I'm not that adept with WPF/XAML, so please let me know particularly if I've done anything wrong or silly regarding the UI.

### DIFF
--- a/POEApi.Infrastructure/DPAPI.cs
+++ b/POEApi.Infrastructure/DPAPI.cs
@@ -12,6 +12,7 @@ namespace POEApi.Infrastructure
     public static class DPAPI
     {
         private static RNGCryptoServiceProvider rngProvider = new RNGCryptoServiceProvider();
+        private static int saltLengthBytes = 16;
 
         public static string Encrypt(this SecureString secret)
         {
@@ -21,7 +22,7 @@ namespace POEApi.Infrastructure
             IntPtr ptr = Marshal.SecureStringToCoTaskMemUnicode(secret);
             try
             {
-                byte[] entropy = new byte[16];
+                byte[] entropy = new byte[saltLengthBytes];
                 rngProvider.GetBytes(entropy);
 
                 char[] buffer = new char[secret.Length];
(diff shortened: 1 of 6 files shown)


thailyn:
Fixed a misleading typo in a comment, and a momentary lapse of "when in Rome" mentality for variable naming.

Stickymaddness:
About the variable naming, I'm not sure what style of variable naming you were using, but I have been thinking for some time that Procurement is _long_ overdue to have the naming brought inline with [modern .NET naming conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines). 

Namely things like,

```c
private int _someInt;
private void SomeMethod();
public const string SomeString = "foo";
```

thailyn:
Yeah, I had initially named `forceRefresh` as `_forceRefresh', as that's the style I'm used to (similar to the excerpt you wrote).

I've wanted to introduce [StyleCop](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md) to this project because the style is a little all-over-the-place, but a) the StyleCop I used in the past has been deprecated, and I need to figure out how to tweak this one, b) I'm not sure of the best way to introduce a style-checker where there are a lot of issues already, c) it's still not a high priority, and d) I don't want to make it like I'm imposing a coding style (it's not my project).

Since you're of a similar mind, though, it seems, I think I should at least file an issue about it and get the ball rolling.

Stickymaddness:
I've wanted to consolidate the style for some time, but it's always been that time spent on Procurement has always gone to features or fixes. 

One approach could be to boyscout naming when making changes or adding new functionality making the trade-off of time spent on naming for consistency across the whole app. 

Either way you have my blessing on this, I'd consider throwing up a naming PR but don't want to cause conflicts with all the good work you've been doing recently unnecessarily.

PR #850, opened 2018-05-27
Fated uniques

thailyn:
Added information about fated uniques to `Data.xml`, and several filters/visitors to find items which relate to these items.

### DIFF
--- a/Procurement/ViewModel/Filters/ForumExport/FatedUniqueBaseTypesFilter.cs
+++ b/Procurement/ViewModel/Filters/ForumExport/FatedUniqueBaseTypesFilter.cs
@@ -1,0 +1,41 @@
+﻿using POEApi.Model;
+using System.Linq;
+
+namespace Procurement.ViewModel.Filters.ForumExport
+{
+    public class FatedUniqueBaseTypesFilter : IFilter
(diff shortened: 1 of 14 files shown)


Procurement/ViewModel/Filters/ForumExport/FatedUniqueBaseTypesFilter.cs
```diff
@@ -0,0 +1,41 @@
...
+            get { return "Unidentified unique items that share a base type with a Fated Unique."; }
+        }
+
+        public bool Applicable(POEApi.Model.Item item)
+        {
+            Gear gear = item as Gear;
+            if (gear == null || gear.Rarity != Rarity.Unique || gear.Identified)
+                return false;
+
+            // TODO: Unique items each have their own, well, unique icon URL.  It is possible to build a map from icon
```
Stickymaddness: A warning on this, urls for the images can change as well, I've seen this happen a few times over the years.

thailyn: I was wondering if that was the case.  Good to know!

PR #888, opened 2018-09-30
Refactor HttpTransport to stop leaking tasks.

thailyn:
(This PR has merge conflicts with #886; I'll fix them when this or that PR is accepted.)

If an HttpWebRequest, such as getting a stash tab, threw an exception, e.g, when getting a 503 response from the server, the HttpTransport would not issue a taskThrottle.CompleteTask() call to match the
taskThrottle.StartTask() from before issuing the request.  If enough requests failed in this way (such as when GGG's servers are fairly unreliable soon after a league starts), the TaskThrottle could "leak" enough running tasks to become completely stuck, such as when refreshing all tabs.

This commit refactors the HttpTransport class so all HttpWebRequests are routed through a single method, which performs the throttled request in a try/finally block.  Thus, if the request fails, the started task is still marked as completed.

There were also several actions which started a task but never indicated it was completed.  This was a "bug", but probably would not ever cause an issue, since they would only ever be done in isolation (such as authenticating the user).  Every action now properly signals its task has finished.

This commit also heavily simplifies the code in the HttpTransport class to simplify it and reduce code duplication.


Stickymaddness:
@thailyn Any chance you can resolve the conflicts with Betrayal drawing nearer?

thailyn:
I've fixed the merge conflicts now with this PR.  I also added a tiny bit more logging that I was working on that will hopefully make debugging issues with the API a lot easier.  We probably should ultimately modify/expand that section (as I indicated in some comments), but it would be excellent if we could at least have the additional logging for the start of the new league, when we have the most issue reports.

Issue #894, opened 2018-12-03
Add support for Betrayal API/item freatures


thailyn:
There are new features in the Betrayal league we can add support for, includes additions to the UI (e.g., veiled items) and new item types (e.g., Scarabs).  This also includes adding the new (revised) set of maps to [`POEApi.Model/Data.xml`](https://github.com/Stickymaddness/Procurement/blob/master/POEApi.Model/Data.xml).

Full details: [link](https://www.pathofexile.com/forum/view-thread/2254801/page/1#p15971585).

Stickymaddness:
How far things have come, API details ahead of time!

thailyn:
Welp, just as we added support for fragment tabs, GGG goes and changes them.  It looks like the only change in what is contained therein is now it includes scarabs.

New background image:

![image](https://user-images.githubusercontent.com/292581/49677643-e834cb80-fa4d-11e8-9338-a11d8af86d4f.png)

aydjay:
Whelp! I'll see where I can get to with what I can find. Could do with a project board of some description!

Stickymaddness:
@aydjay By project board are you referring to Betrayal changes in general? If so I can think of a few things off the top of my head that still need to be implemented,

- [ ] Fragment tabs changes
- [ ] (Un)Veiled mod support
- [ ] [Header](https://web.poecdn.com/image/layout/header.png?1544131088000) update
- [ ] Procurement is also still missing map tab support (this one will be a PITA).

aydjay:
Yep - that's exactly what I was thinking off.

Stickymaddness:
Closing this as we managed to add all the missing functionality. Lets see what the new league announcement brings us next week!

PR #895, opened 2018-12-04
Add new divination cards in 3.5

Stickymaddness:
#894


thailyn:
Any chance we could keep the list alphabetized? 👼

Stickymaddness:
Good catch, #896

PR #903, opened 2018-12-14
Leverage Cloud flare utilities to login

aydjay:
This is my submission to avoid cloud flare DDoS protection

This doesn't need the extra data that @thailyn's approach uses and offloads the cloudflare bypass off to a 3rd party nuget package:

https://github.com/elcattivo/CloudFlareUtilities

Projects had to be upgraded to .net 4.5 to accomodate the .Net standard CloudFlareUtilities package.

Feel free to reject if @thailyn's approach is preferred.

image

### DIFF
--- a/POEApi.Transport/HttpTransport.cs
+++ b/POEApi.Transport/HttpTransport.cs
@@ -86,23 +86,15 @@ namespace POEApi.Transport
                     TraditionalSessionIdLogin();
                     return true;
                 }
-                catch (Exception ex)
+                catch (WebException ex)
                 {
-                    //Something that isn't a cloud flare exception occurred
-                    if (ex.Message.Contains("503") == false)
-                    {
-                        throw;
-                    }
-
-                    try
+                    if (((HttpWebResponse)ex.Response).Server == "cloudflare" && ((HttpWebResponse)ex.Response).StatusCode == HttpStatusCode.ServiceUnavailable)
                     {
                         CloudFlareSessionIdLogin();
                         return true;
                     }
-                    catch
-                    {
-                        throw;
-                    }
+
+                    throw;
                 }
             }
 
@@ -144,7 +136,7 @@ namespace POEApi.Transport
             catch (AggregateException ex) when (ex.InnerException is CloudFlareClearanceException)
             {
                 // After all retries, clearance still failed.
-                throw new Exception("Cloud flare clearance failed, please wait one minute and try again");
+                throw new Exception("Cloud flare clearance failed, please wait one minute and try again", ex);
             }
             catch (AggregateException ex) when (ex.InnerException is TaskCanceledException)
             {
--- a/POEApi.Transport/HttpTransport.cs
+++ b/POEApi.Transport/HttpTransport.cs
@@ -1,11 +1,14 @@
 ﻿using System;
 using System.IO;
 using System.Net;
+using System.Net.Http;
 using System.Text;
 using POEApi.Infrastructure;
 using System.Security;
 using POEApi.Infrastructure.Events;
 using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using CloudFlareUtilities;
 
 namespace POEApi.Transport
 {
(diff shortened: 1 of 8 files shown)


POEApi.Transport/HttpTransport.cs
```diff
@@ -77,13 +80,30 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
+                    {
+                        throw;
+                    }
+
+                    try
+                    {
+                        CloudFlareSessionIdLogin();
+                        return true;
+                    }
+                    catch
```
thailyn: Since we're just rethrowing any exceptions, this try/catch block around this is useless.

POEApi.Transport/HttpTransport.cs
```diff
@@ -77,13 +80,30 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
+
+                try
                 {
-                    // If the response URI is the login URL, then the login action failed.
-                    if (sessionIdLoginResponse.ResponseUri.ToString() == loginURL)
-                        throw new LogonFailedException();
+                    TraditionalSessionIdLogin();
+                    return true;
+                }
+                catch (Exception ex)
```
thailyn: Specifically catch a `System.Net.WebException` instead.

POEApi.Transport/HttpTransport.cs
```diff
@@ -77,13 +80,30 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
+                {
+                    //Something that isn't a cloud flare exception occurred
+                    if (ex.Message.Contains("503") == false)
```
thailyn: When we're catching specifically a `WebException` and an issue with CloudFlare happens, we can specifically check for:

`((System.Net.HttpWebResponse)ex.Response).Server == "cloudflare"`

and that `((HttpWebResponse)ex.Response).StatusCode` has the enum value `ServiceUnavailable`

POEApi.Transport/HttpTransport.cs
```diff
@@ -106,6 +126,43 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
+        private void CloudFlareSessionIdLogin()
+        {
+            try
+            {
+                using (var clearanceHandler = new ClearanceHandler {MaxRetries = 2})
+                using (var handler = new HttpClientHandler() {CookieContainer = credentialCookies, Proxy = getProxySettings()})
+                {
+                    clearanceHandler.InnerHandler = handler;
+
+                    using (var client = new HttpClient(clearanceHandler))
```
thailyn: This is not the correct way to be using `HttpClient` (creating a new copy for every request; see [here](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/)).  **However,** since by design we're only using the `HttpClient` once per every time we run Procurement, it is okay.  When we switch to only using `HttpClient`, we'll need to be more thorough.

POEApi.Transport/HttpTransport.cs
```diff
@@ -106,6 +126,43 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
+                    {
+                        var result = client.GetStringAsync(new Uri(loginURL)).Result;
+                    }
+                }
+            }
+            catch (AggregateException ex) when (ex.InnerException is CloudFlareClearanceException)
+            {
+                // After all retries, clearance still failed.
+                throw new Exception("Cloud flare clearance failed, please wait one minute and try again");
```
thailyn: Include the `AggregateException ex` as the second parameter to the `Exception` constructor, so it and its details are not lost.

POEApi.Transport/HttpTransport.cs
```diff
@@ -106,6 +126,43 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
         }
 
+        private void CloudFlareSessionIdLogin()
+        {
+            try
+            {
+                using (var clearanceHandler = new ClearanceHandler {MaxRetries = 2})
+                using (var handler = new HttpClientHandler() {CookieContainer = credentialCookies, Proxy = getProxySettings()})
+                {
+                    clearanceHandler.InnerHandler = handler;
```
thailyn: So that's how we can use multiple handlers!  I was trying to figure that out!

POEApi.Transport/HttpTransport.cs
aydjay: Great spot!

POEApi.Transport/HttpTransport.cs
aydjay: I wanted to leave as much alone as possible xD

aydjay:
Acted on your comments @thailyn 

https://github.com/Stickymaddness/Procurement/blob/51e5239ef38748f6979abf9b2e0589cddff9b135/POEApi.Transport/HttpTransport.cs#L91

I was looking for a better way, this is very much better, thanks.

aydjay:
PR is good to go from my perspective

POEApi.Transport/HttpTransport.cs
```diff
@@ -86,23 +86,15 @@ public bool Authenticate(string email, SecureString password, bool useSessionID)
...
+                catch (WebException ex)
                 {
-                    //Something that isn't a cloud flare exception occurred
-                    if (ex.Message.Contains("503") == false)
-                    {
-                        throw;
-                    }
-
-                    try
+                    if (((HttpWebResponse)ex.Response).Server == "cloudflare" && ((HttpWebResponse)ex.Response).StatusCode == HttpStatusCode.ServiceUnavailable)
```
thailyn: Nit: split this condition across two lines (too long right now).

aydjay: I'm away from my development machine until after the weekend. I'll leave this for now.

Stickymaddness:
Great work @aydjay, you and @thailyn are really carrying this project!

The .NET 4.5 upgrade is long overdue, so I'm fine with bumping it. There's a lot we can do with the HttpClient handling, not to mention a big refactor and cleanup of things.

PR #905, opened 2018-12-17
Tab refinements

aydjay:
There are a number of refinements in this PR.

Please forgive me.

Condense all Commands to be ICommand and Leverage RelayCommand. Using the implementation of Relay command by Josh Smith: https://joshsmithonwpf.wordpress.com/

Condense all implementations of INPC into a base class (ObservableBase) and let all the viewmodels inherit it.

Give premium tabs their coloured border inside the application (Ultra minor but it's been annoying me for some time)

Update the Fragment stash tab background and positions

Add the Scarabs to the Fragment Stash tab viewmodel.

Before I continue with the scarabs I need some more info @thailyn:

Rank placements:
Rusted, Polished, Gilded (Left to right)
Effect Placements:
Which effect where!

If anyone has a fragment tab with some of each grade of scarab i would be very appreciative as a) I don't have a fragment stash tab and b) I don't have a single scarab yet! (Still on lvl 50 on one char)

@Stickymaddness Please merge if you see fit, otherwise I will hold off until all the scarabs are neatly placed on the fragment stash.


thailyn:
It's awesome that you've gotten all of this done, but I'm wondering if it
would be possible to split this into multiple PRs, possibly four or so?
Especially since they're split up by commits already.  Having smaller PRs
make them easier to review, and keeps the commit history cleaner (commits
refining a feature during review, if necessary, are close to the original
commit and not interspersed with unrelated commits).

Also I haven't gotten any scarabs yet in game (just did my first safehouse
yesterday -- so slow! >_<), so I might need a bit of time to do some
digging and figure out the placement and other properties.

aydjay:
I'll close and re-open smaller PR's

PR #907, opened 2018-12-18
Relay command

aydjay:
PR based on #906, Please merge that one first.

Condense all Commands to be ICommand and Leverage RelayCommand. Using the implementation of Relay command by Josh Smith: https://joshsmithonwpf.wordpress.com/

This will allow mocking of the command behaviours in the future in an tests.


Procurement/RelayCommand.cs
```diff
@@ -0,0 +1,54 @@
+using System;
+using System.Diagnostics;
+using System.Windows.Input;
+
+namespace Procurement.View.ViewModel
+{
+    public class RelayCommand : ICommand
+    {
+        #region Fields
```
Stickymaddness: I'd like to avoid adding regions, doubly so in a small class like this.

aydjay: Agreed.

PR #909, opened 2018-12-18
Scarab fragments

aydjay:
Update the Fragment stash tab background and positions

Add the Scarabs to the Fragment Stash tab viewmodel.

Before I continue with the scarabs I need some more info @thailyn:

Rank placements:
Rusted, Polished, Gilded (Left to right)
Effect Placements:
Which effect where!

If anyone has a fragment tab with some of each grade of scarab i would be very appreciative as:
* I don't have a fragment stash tab
* I don't have a single scarab yet! (Still on lvl 50 on one char)

Needs #908 merging first


Stickymaddness:
> CurrencyStash.xaml.cs(49,103): error CS1061: 'Colour' does not contain a definition for 'WpfColor' and no extension method 'WpfColor' accepting a first argument of type 'Colour' could be found (are you missing a using directive or an assembly refe

aydjay:
Looks like I will have to revisit this until other PR's are merged in.

aydjay:
Closing this because i've fucked up. Git noob here....

PR #921, opened 2018-12-22
Add support for veiled mods.

thailyn:
This PR adds veiled mods as a property of `Item` objects, alongside crafted mods (since veiled mods become crafted mods after being unveiled), and displays a simple text string in the item hover.  We can change the text string later to be similar to the animation that is currently shown in game, but for now there is some indication that veiled mods are present.

### DIFF
--- a/POEApi.Model/Item.cs
+++ b/POEApi.Model/Item.cs
@@ -53,6 +53,7 @@ namespace POEApi.Model
         public List<string> FlavourText { get; set; }
 
         public List<string> CraftedMods { get; set; }
+        public List<string> VeiledMods { get; set; }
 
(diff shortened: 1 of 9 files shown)


Stickymaddness:
Nice one, didn't realise there was more coming! Just put a release out seconds ago ;)

Would be nice to add some tests around veiled items.

thailyn:
Yeah, I was delayed a bit because I was trying to figure out how the veiled mod graphics worked.  It's a CSS animation sliding across a background on the web page, but the WPF knowledge to do something similar is currently beyond me.  I'll create a feature request documenting what I found shortly.

Issue #929, opened 2018-12-31
Support 21-23% quality gems


fqrwx:
When I set price to one of these gems, they are not included in the forum code, I had to manually add them

thailyn:
Under which category are they not showing up/should they show up in?

fqrwx:
I believe they need to be supported in the template like this one but for 21 22 and 23:

```
[spoiler="          Quality Gems 15-20    "]
{Quality16Gems}{Quality17Gems}{Quality18Gems}{Quality19Gems}{Quality20Gems}
[/spoiler]
```

fqrwx:
found some minor issue with the gem spoiler titles too

5-10, 10-15, 15-20 needs to be replaced with 6-10, 11-15 and 16-20 respectively

fqrwx:
is there a {Tier16Maps} needed for tier 16 maps also?

aydjay:
Closed by #932

Issue #925, opened 2018-12-26
Scarabs missing from template


MajesticPeter:
Scarabs don't get added to the forum post since they're missing from the template. I'm using regular item storage tabs (no premium tabs).

thailyn:
Hi, you can add `{Scarab}` to the template and it should show your scarabs in the forum post.

PR #932, opened 2019-01-03
Update jeweller's tag and add missing items to forum export

### LINKED ISSUE #928: Update jewellers tag
The items that have their prices set to jewellers in Procurement don't show up in searches, they have no price shown, I think the tag needs to be jew instead of jewel. I would like to see more currency types supported too (https://www.pathofexile.com/trade/about, http://currency.poe.trade/tags)

fqrwx:
Should fix #928 and #929 however untested and something might be missing or wrong


fqrwx:
Added scarabs for #925 unless that feature isn't ready yet?

PR #933, opened 2019-01-04
Support Vaal Orb Recipe

### LINKED ISSUE #927: Vaal orb recipe
7 x vaal skill gems and a sac fragment (Leave the midnight as a lower priority)

image

aydjay:
Added Vaal Orb recipe close #927 

I've noticed that the item highlighting once you've selected a recipe component is no longer working. This is probably a side from #915

### DIFF
--- a/Procurement/Procurement.csproj
+++ b/Procurement/Procurement.csproj
@@ -227,6 +227,7 @@
     <Compile Include="ViewModel\Recipes\MatchedSet.cs" />
     <Compile Include="ViewModel\SetTabBuyoutViewModel.cs" />
     <Compile Include="ViewModel\TabViewModel\CommonTabViewModel.cs" />
+    <Compile Include="ViewModel\TabViewModel\TabFactory.cs" />
     <Compile Include="ViewModel\TradeSettingsViewModel.cs" />
     <Compile Include="ViewModel\Cache\BitmapCache.cs" />
     <Compile Include="ViewModel\DisplayModeStrategy\DisplayModeFactory.cs" />
--- a/Procurement/Procurement.csproj
+++ b/Procurement/Procurement.csproj
@@ -109,6 +109,7 @@
     <Compile Include="Controls\FragmentStashTab.xaml.cs">
       <DependentUpon>FragmentStashTab.xaml</DependentUpon>
     </Compile>
+    <Compile Include="ViewModel\Recipes\ChromaticRecipe.cs" />
     <Compile Include="ViewModel\TabViewModel\FragmentStashViewModel.cs" />
     <Compile Include="RelayCommand.cs" />
     <Compile Include="ViewModel\TabViewModel\EssenceStashViewModel.cs" />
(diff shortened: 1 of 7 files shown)


aydjay:
Bumping for a merge please

Procurement/Procurement.csproj
```diff
@@ -109,6 +109,7 @@
     <Compile Include="Controls\FragmentStashTab.xaml.cs">
       <DependentUpon>FragmentStashTab.xaml</DependentUpon>
     </Compile>
+    <Compile Include="ViewModel\Recipes\ChromaticRecipe.cs" />
```
thailyn: I know the list of files here isn't sorted and is primarily managed by Visual Studio, but maybe manually move this line back to where it was, so we don't make the sorting worse.

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
```diff
@@ -0,0 +1,67 @@
...
+                    Instance = this,
+                    MatchedItems = new List<Item>(),
+                    Missing = new List<string>(),
+                    IsMatch = true
+                };
+
+                for (int i = candidateGems.Count - 1; i >= 0; i--)
+                {
+                    var candidateGem = candidateGems[i];
+                    if (recipeResult.MatchedItems.Count <= 6)
```
thailyn: Instead of using magic numbers `6` and `7`, create a class member representing the required number of gems, and use that where required.

thailyn: If you invert this `if` condition, you can make the code a little simpler (no `else` case).  You can also define `candidateGem` after the `if` block, to right before it's used.

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
```diff
@@ -0,0 +1,67 @@
...
+                    else
+                    {
+                        break;
+                    }
+                }
+
+                var numberOfMissingGems = 7 - recipeResult.MatchedItems.Count;
+                if (numberOfMissingGems >= 1)
+                {
+                    recipeResult.Missing.Add($"{numberOfMissingGems} Vaal Skill gems");
```
thailyn: Handle the case where only one gem is missing and the text here should read, "1 Vaal Skill gem".

Procurement/Procurement.csproj
aydjay: I let VS handle the project files. I don't mess with them :)

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
aydjay: Added constant and changed logic so that only 7 is needed.

aydjay: Done - idk why I didn't do it in the first place...

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
aydjay: Good idea.

aydjay:
All comments resolved - ready for a merge.

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
```diff
@@ -57,7 +55,7 @@ public override IEnumerable<RecipeResult> Matches(IEnumerable<Item> items)
                     recipeResult.Missing.Add("Sacrifice Fragment");
                 }
 
-                recipeResult.PercentMatch = (decimal) ((recipeResult.MatchedItems.Count / 8d) * 100) ;
+                recipeResult.PercentMatch = (decimal) (recipeResult.MatchedItems.Count / 8d * 100) ;
```
thailyn: `NeededVaalGems + 1d` instead of `8d` (or even also add a member for the number of needed sacrifice fragments).

Stickymaddness:
@aydjay Please rebase as @thailyn has suggested to clean up the commit history, and we can get this merged.

aydjay:
I have no idea what's gone wrong with this PR. I need to understand a bit more about rebasing - I'm clearly doing something wrong. I'm going to close this and re-open a fresh PR.

PR #934, opened 2019-01-04
Allow highlights to apply across all stash tab types

aydjay:
I saw that recipes no longer highlight while working on the vaal orb recipe after the #915 

This is kind of a key feature of procurement!

I've remedied this and allowed the filter to apply to any type of stash tab.


aydjay:
Bumping for a merge please

PR #938, opened 2019-01-29
Support Vaal Orb Recipe

### LINKED ISSUE #927: Vaal orb recipe

aydjay:
image

I closed #927  as the commit history had gotten out of hand.

### DIFF
--- a/Procurement/ViewModel/Recipes/RecipeManager.cs
+++ b/Procurement/ViewModel/Recipes/RecipeManager.cs
@@ -28,7 +28,7 @@ namespace Procurement.ViewModel.Recipes
                 new SameBaseTypeRecipe(),
                 new SameNameRecipe("Alchemy Orb - 3 Of The Same Name", 3),
                 new SameNameRecipe("Chance Orb - 2 Of The Same Name", 2),
-                new VaalOrbRecipe()
+                new VaalOrbRecipe(),
                 //Todo: Implement Essence Combination recipe (Exclude Shrieking and "Special" essences
             };
         }
--- a/Procurement/ViewModel/Recipes/RecipeManager.cs
+++ b/Procurement/ViewModel/Recipes/RecipeManager.cs
@@ -28,6 +28,7 @@ namespace Procurement.ViewModel.Recipes
                 new SameBaseTypeRecipe(),
                 new SameNameRecipe("Alchemy Orb - 3 Of The Same Name", 3),
                 new SameNameRecipe("Chance Orb - 2 Of The Same Name", 2),
+                new VaalOrbRecipe()
                 //Todo: Implement Essence Combination recipe (Exclude Shrieking and "Special" essences
             };
         }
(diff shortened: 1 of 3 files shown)


Procurement/ViewModel/Recipes/RecipeManager.cs
```diff
@@ -28,6 +28,7 @@ public RecipeManager()
                 new SameBaseTypeRecipe(),
                 new SameNameRecipe("Alchemy Orb - 3 Of The Same Name", 3),
                 new SameNameRecipe("Chance Orb - 2 Of The Same Name", 2),
+                new VaalOrbRecipe()
```
thailyn: Style issue, but I would recommend adding a trailing comma to this line, to minimize the number of changed lines when another recipe is added.

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                    recipeResult.MatchedItems.Add(candidateGem);
+                    candidateGems.Remove(candidateGem);
+                }
+
+                var numberOfMissingGems = NeededVaalGems - recipeResult.MatchedItems.Count;
+
+                if (numberOfMissingGems == 1)
+                {
+                    recipeResult.Missing.Add($"{numberOfMissingGems} Vaal Skill gem");
```
thailyn: Maybe the compiler optimizes it away anyway, but since we know here `numberOfMissingGems` is 1, we can hard-code it into the string literal, which would be (perhaps trivially) faster.

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    recipeResult.MatchedItems.Add(fragments.First());
+                    fragments.RemoveAt(0);
+                }
+                else
+                {
+                    recipeResult.Missing.Add("Sacrifice Fragment");
+                }
+
+                recipeResult.PercentMatch = (decimal) (recipeResult.MatchedItems.Count / 8d * 100) ;
```
thailyn: `NeededVaalGems` + 1d instead of 8d (or even also add a member for the number of needed sacrifice fragments).

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
aydjay: Good spot

Procurement/ViewModel/Recipes/RecipeManager.cs
aydjay: Minor but I like to keep you happy xD

Procurement/ViewModel/Recipes/VaalOrbRecipe.cs
aydjay: I added a decimal constant for the total number of needed items, saves casting and computations etc...

Issue #819, opened 2018-03-12
Exalt shard recipe


aydjay:
https://www.reddit.com/r/pathofexile/comments/83s0o2/a_few_more_recipes/

Full set of elder/shaper sets id'd == 2 exalt shards
Full set of elder/shaper sets unid'd == 4 exalt shards

PR #940, opened 2019-01-29
Support Exalted Shard Recipe

aydjay:
image

Minor adjustments to the rare set recipe to accommodate the shaper/elder set recipe.

Closes: #819

### DIFF
--- a/Procurement/ViewModel/Recipes/RareSetRecipe.cs
+++ b/Procurement/ViewModel/Recipes/RareSetRecipe.cs
@@ -40,19 +40,10 @@ namespace Procurement.ViewModel.Recipes
         public override IEnumerable<RecipeResult> Matches(IEnumerable<Item> items)
         {
             List<Gear> allGear = items.OfType<Gear>().ToList();
-            bool isShaperWanted = false;
-            bool isElderWanted = false;
 
-            switch (_setType)
-            {
-                case SetType.Shaper:
-                    isShaperWanted = true;
-                    break;
-                case SetType.Elder:
-                    isElderWanted = true;
-                    break;
-            }
-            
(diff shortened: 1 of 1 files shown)


Procurement/ViewModel/Recipes/RareSetRecipe.cs
```diff
@@ -19,23 +27,39 @@ public override string Name
...
             this.maximumItemLevel = maximumItemLevel;
             this.itemsIdentified = itemsIdentified;
             this.name = name;
+            _setType = setType;
         }
 
         public override IEnumerable<RecipeResult> Matches(IEnumerable<Item> items)
         {
             List<Gear> allGear = items.OfType<Gear>().ToList();
+            bool isShaperWanted = false;
```
thailyn: You could do something like the following and not need the switch statement.

``` C#
bool isShaperWanted = _setType == SetType.Shaper;
bool isElderWanted = _setType == SetType.Elder;
```

Alternatively, since these variables are only being used in one place, you could instead check them in the `Where` function, below.  But that might make the already-complicated `Where` function overly complex.

aydjay: I knew there was a better way! I had considered adding to the LINQ but it's already complex enough.

Procurement/ViewModel/Recipes/RareSetRecipe.cs
```diff
@@ -5,12 +5,20 @@
 
 namespace Procurement.ViewModel.Recipes
 {
+    public enum SetType
+    {
+        Normal,
+        Shaper,
+        Elder
```
thailyn: As in the other PR, I'd recommend adding a trailing comma here, but, again, it's a minor style nitpick.

PR #942, opened 2019-01-29
Memory management

aydjay:
As I was clicking between the stash view and the recipe view I noticed that it was a smidge slow displaying the stash. When I looked at the resource manager I also noticed that the memory held by procurement was spiking pretty hard.

I dug through the main window code and saw that there was a lack of binding and quite a bit of "Extra" work being done that isn't needed.

The stash view is now being regenerated only on a refresh and we use data bindings to show the view that we want. There is something in the "StashVIew" that isn't releasing cleanly so persistent refreshes will still cause the memory to climb but this PR irons out some of the issue.


aydjay:
This is ready for a merge.

Quite a few issues with event handlers not getting released properly in ItemDisplay.xaml.cs

PR #944, opened 2019-01-31
Display gem progress in ItemHover

aydjay:
Fairly simple change, add gem experience to item hover.

### DIFF
--- a/POEApi.Model/Gem.cs
+++ b/POEApi.Model/Gem.cs
@@ -6,6 +6,9 @@ namespace POEApi.Model
     public class Gem : SocketableItem
     {
         public int Level { get; set; }
+        public double LevelProgress { get; private set; }
+        public int Numerator { get; set; }
+        public int Denominator { get; set; }
 
         public Gem(JSONProxy.Item item) : base(item)
         {
@@ -13,6 +16,31 @@ namespace POEApi.Model
             this.ItemType = ItemType.Gem;
 
             this.Level = getLevel();
+
+            ExtractGemExperience(item);
+        }
+
+        private void ExtractGemExperience(JSONProxy.Item item)
+        {
+            var experienceProperties = item.AdditionalProperties.FirstOrDefault(x => x.Name == "Experience");
+
+            if (experienceProperties == null)
+            {
+                return;
+            }
+
+            LevelProgress = experienceProperties.Progress;
+            var experienceValues = experienceProperties.Values[0].FirstOrDefault()?.ToString();
+
+            if (experienceValues == null || !experienceValues.Contains("/"))
+            {
+                return;
+            }
+
+            var numeratorAndDenominator = experienceValues.Split('/');
+
+            Numerator = int.Parse(numeratorAndDenominator[0]);
+            Denominator = int.Parse(numeratorAndDenominator[1]);
         }
 
         private int getLevel()
--- a/POEApi.Model/Gem.cs
+++ b/POEApi.Model/Gem.cs
@@ -6,9 +6,21 @@ namespace POEApi.Model
     public class Gem : SocketableItem
     {
         public int Level { get; set; }
-        public double LevelProgress { get; private set; }
-        public int Numerator { get; set; }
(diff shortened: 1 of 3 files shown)


POEApi.Model/Gem.cs
```diff
@@ -6,13 +6,41 @@ namespace POEApi.Model
     public class Gem : SocketableItem
     {
         public int Level { get; set; }
+        public double LevelProgress { get; private set; }
```
thailyn: Add a comment explaining what `LevelProgress` is.  If I remember correctly from the API, it's the fraction the gem has progressed to the next level? (`LevelExperienceProgress` might be a better, more explicit name, then.)

POEApi.Model/Gem.cs
```diff
@@ -6,13 +6,41 @@ namespace POEApi.Model
...
+        public int Numerator { get; set; }
```
thailyn: Update the names here to reflect that they are the numerator and denominator that relate to the gem's current experience.  Adding comments with more detail wouldn't hurt, either.

POEApi.Model/Gem.cs
```diff
@@ -6,13 +6,41 @@ namespace POEApi.Model
...
+            var experienceValues = experienceProperties.Values[0].FirstOrDefault()?.ToString();
+
+            if (experienceValues == null || !experienceValues.Contains("/"))
+            {
+                return;
+            }
+
+            var numeratorAndDenominator = experienceValues.Split('/');
+
+            Numerator = int.Parse(numeratorAndDenominator[0]);
```
thailyn: Perhaps use something like `int.TryParse`?  If we're already assuming `numeratorAndDenominator` might be null or not have a `/` character, we might as well assume it might not have valid numbers.  I assume not, but just to double check: is it possible for the numerator to not be a whole number?

Procurement/Controls/ItemHover.xaml
```diff
@@ -423,17 +432,40 @@
...
                                                     Converter={StaticResource bc},
                                                     ConverterParameter=CollapseWhenFalse}" />
 
+                    <StackPanel Orientation="Vertical"
+                                Visibility="{Binding IsGemProgressVisible,
+                                                 Converter={StaticResource VisibilityConverter},
+                                                 ConverterParameter=CollapseWhenFalse}">
+
+                        <DockPanel Margin="10 0 10 5">
+
```
thailyn: Remove the two blank lines here, and the one on line 450.

Procurement/ViewModel/ItemHoverViewModel.cs
```diff
@@ -44,6 +44,13 @@ public class ItemHoverViewModel
 
         public string ItemLevel { get; set; }
 
+        public bool IsGemProgressVisible => Item is Gem;
```
thailyn: What if the Json did not have valid experience information?  How would the progress bar be displayed then?

Procurement/ViewModel/ItemHoverViewModel.cs
```diff
@@ -44,6 +44,13 @@ public class ItemHoverViewModel
...
+
+        public double LevelProgress { get; set; }
```
thailyn: Same comments here about the variable names and comments for `LevelProgress`, `Numerator`, and `Denominator`.

Procurement/ViewModel/ItemHoverViewModel.cs
aydjay: This is what the fallback values are for in the ItemHover.xaml

If the item isn't a gem the whole section won't be visible.

Procurement/Controls/ItemHover.xaml
aydjay: Done

Procurement/ViewModel/ItemHoverViewModel.cs
aydjay: Made them match up with the variable on the gem.

PR #945, opened 2019-02-01
Lined up currencies

aydjay:
Very minor but it was annoying me

Old:

New:

Feel free to reject if it's not an undesired change.


thailyn:
Just looking at the preview, I have two thoughts:
* Can you right-align the currency values?  Then the decimal places would line up and it would be easier to compare values.
* Moving the values off to the right makes it a little harder to see the value of an individual currency, specifically when not hovering over the currency with the mouse (which is essentially all the time, for me).  Can you add some UI feature to make it easier for the eye to move between the two columns (e.g., keep the mouse-over effect permanently, or add grid lines)?

aydjay:
I had pondered about right aligning the currency values, I will do it.

Agreed on the on hover effect, I'll add grid lines in.

PR #948, opened 2019-02-05
Update list of known divination cards

aydjay:
Source data: https://pathofexile.gamepedia.com/Divination_card

Some cards are not in the source data as they have had their drops restricted but they could exist in standard so we need to keep them in the list.

### DIFF
--- a/POEApi.Model/Data.xml
+++ b/POEApi.Model/Data.xml
@@ -1485,6 +1485,9 @@
       <Item name="Profane Wand"/>
     </GearBaseType>
 
+    <!--Source data: https://pathofexile.gamepedia.com/Divination_card-->
+    <!--Regex find: ^(.*)\r -->
+    <!--Regex Replace: <Item name="$1"/> -->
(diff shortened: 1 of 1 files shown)


POEApi.Model/Data.xml
```diff
@@ -1694,10 +1704,11 @@
       <Item name="The Warlord"/>
       <Item name="The Watcher"/>
       <Item name="The Web"/>
+      <Item name="The Wilted Rose"/>
       <Item name="The Wind"/>
       <Item name="The Witch"/>
-      <Item name="The Wolf"/>
       <Item name="The Wolf's Shadow"/>
+      <Item name="The Wolf"/>
```
thailyn: Nit: You're sorting this after `The Wolf's Shadow`, but `The Master Artisan` is being moved to before `The Master`, which doesn't seem consistent.  I usually order by "nothing before something", so I would expect `The Wolf` to sort before `The Wolf's Shadow`.

Issue #151, opened 2015-08-04
[Feature Request] Copy item info to clipboard via ^c similar to in-game


trackpete:
I'm working on a new version of my price macro that can do rudimentary matching for rares and other items and works by taking a [full item copy/paste submission for the parsing](http://exiletools.com/price-lookup-test.html) (so it's all done server side). 

I was thinking that if you could add similar text output functionality to Procurement on mouse-over and ^c, I could modify the macro to also work within Procurement. It doesn't have to be a perfect clone of in-game (obviously it won't have Item Level/etc.), but just the basic layout as already shown by the javascript renderer would be fantastic. 

Right now when I'm setting prices I have to have an in-game window open to run the macro on the items, but it'd be great to be able to do it straight from Procurement. Bonus: This wouldn't require you to actually interface with my price macro/upcoming API.

Cheers,

 -Pete

Stickymaddness:
Hi Pete,

I'm still very keen to add this functionality to Procurement, the previous pull request kind of fell by the wayside. What is the current state of your API? 

Do you have an endpoint I can call with some rudimentary item data that will bring back pricing info? As you mentioned, the only limitation should be item level.

trackpete:
I am currently in development on a whole new system, however you can take a look at an early version (I am really hoping to push this out in the next couple weeks), linked above. It's really simple, to hit the dev version basically send a POST request to http://api.exiletools.com/cgi-bin/dev-item-report-text with the following:

v=5
league=[league - warbands|tempest|hardcore|softcore]
itemData=[url encoded text data for an item as seen in-game]

There's also a fully functional version available at api.exiletools.com/item-report-text that works the same way, but it currently only operates on Gems and Uniques.

Right now the data returned is purely text output, intended for display in a popup. I will be adding a JSON output format in the near future, but right now I'm working hard to get my indexer handling all those stupid Acquisition threads that aren't getting bumped. ;(

tl;dr: old price API is completely gone, but the new one is way better, just need to finish work on it, which will hopefully be very soon.

Stickymaddness:
I tried pasting text for rares in http://exiletools.com/price-lookup-test.html and making a post request to http://api.exiletools.com/cgi-bin/dev-item-report-text, but I get the following for any items I try:

Sorry! 0 out of 1420533 verified items with prices
updated in the last 7 days matched this search.

Tested with rare Sharkskin Gloves and Two-Stone ring.

trackpete:
[...]

PR #954, opened 2019-02-07
Add ability to export items into pob compatible format

aydjay:
Closes: #151 

I've done a fair amount of testing and the only anomaly I discovered was sacrifice fragments getting picked up. The code generates successfully but pob doesn't accept it. This is because the fragments are getting detected as "Gear" which is incorrect but is out of the scope of this PR to resolve.

### DIFF
--- a/POEApi.Model/Gear.cs
+++ b/POEApi.Model/Gear.cs
@@ -86,7 +86,21 @@ namespace POEApi.Model
 
                     if (Sockets != null && Sockets.Any())
                     {
-                        var socketData = string.Join("=", Sockets.Select(x => x.ToPobFormat()));
+                        //Links are denoted by "Groups"
+                        List<int> groups = Sockets.Select(x => x.Group).Distinct().ToList();
+                        string socketData = string.Empty;
+                            
+                        foreach (var group in groups)
+                        {
+                            socketData += string.Join("=", Sockets.Where(x => x.Group == group).Select(y => y.ToPobFormat()));
+                            
+                            //Don't append a space on the last character
+                            if (groups.Last() != group)
+                            {
+                                //Space character denotes group demarcations
+                                socketData += " ";
+                            }
+                        }
 
                         pobData.AppendLine($"Sockets: {socketData}");
                     }
--- a/POEApi.Model/Gear.cs
+++ b/POEApi.Model/Gear.cs
@@ -1,5 +1,8 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
+using System.Text;
+using POEApi.Infrastructure;
 
 namespace POEApi.Model
 {
@@ -65,6 +68,52 @@ namespace POEApi.Model
                 }
 
                 return components;
+            }
+        }
+
+        public override string PobData
+        {
+            get
+            {
+                try
+                {
+                    var pobData = new StringBuilder();
+                    pobData.AppendLine(Name);
+                    pobData.AppendLine(TypeLine);
+                    pobData.AppendLine($"Unique ID: {Id}");
+                    pobData.AppendLine($"Item Level: {ItemLevel}");
+                    pobData.AppendLine($"Quality: {Quality}");
+
+                    if (Sockets != null && Sockets.Any())
+                    {
+                        var socketData = string.Join("=", Sockets.Select(x => x.ToPobFormat()));
+
+                        pobData.AppendLine($"Sockets: {socketData}");
+                    }
+
+                    if (Implicitmods != null && Implicitmods.Any())
+                    {
+                        pobData.AppendLine($"Implicits: {Implicitmods.Count}");
+                        Implicitmods.ForEach(x => pobData.AppendLine($"{{crafted}}{x}"));
+                    }
+
+                    if (Explicitmods != null && Explicitmods.Any())
+                    {
+                        Explicitmods.ForEach(x=> pobData.AppendLine(x));
+                    }
+
+                    if (CraftedMods != null && CraftedMods.Any())
+                    {
+                        CraftedMods.ForEach(x => pobData.AppendLine($"{{crafted}}{x}"));
+                    }
+
+                    return pobData.ToString();
+                }
+                catch (Exception e)
+                {
+                    Logger.Log(e);
+                    throw;
+                }
             }
         }
     }
--- a/POEApi.Model/Socket.cs
+++ b/POEApi.Model/Socket.cs
@@ -20,17 +20,22 @@
             switch (Attribute)
             {
                 case "S":
-                    return "R";
+                    return RED;
                 case "I":
-                    return "B";
+                    return BLUE;
                 case "D":
-                    return "G";
+                    return GREEN;
                 case "G":
-                    return "W";
+                    return WHITE;
                 default:
-                    //Abyssal socket.
-                    return "A"; 
+                    return ABYSSAL; 
             }
         }
+
+        private const string RED = "R";
+        private const string BLUE = "B";
+        private const string GREEN = "G";
+        private const string WHITE = "W";
+        private const string ABYSSAL = "A";
     }
 }
--- a/POEApi.Model/Socket.cs
+++ b/POEApi.Model/Socket.cs
@@ -10,5 +10,27 @@
             this.Attribute = s.Attr;
             this.Group = s.Group;
         }
+
+        /// <summary>
+        /// Convert the Socket attribute to something consumable by path of building
+        /// </summary>
+        /// <returns></returns>
+        public string ToPobFormat()
+        {
+            switch (Attribute)
+            {
+                case "S":
+                    return "R";
+                case "I":
+                    return "B";
+                case "D":
+                    return "G";
+                case "G":
+                    return "W";
+                default:
+                    //Abyssal socket.
+                    return "A"; 
+            }
+        }
     }
 }
--- a/Procurement/Controls/ItemDisplay.xaml.cs
+++ b/Procurement/Controls/ItemDisplay.xaml.cs
@@ -196,7 +196,7 @@ namespace Procurement.Controls
             var setBuyout = new MenuItem();
             setBuyout.StaysOpenOnClick = true;
 
-            var buyoutControl = new SetBuyoutView();
+            var buyoutControl = new SetBuyoutView(item);
 
             if (Settings.Buyouts.ContainsKey(item.Id))
             {
--- a/Procurement/Controls/ItemDisplay.xaml.cs
+++ b/Procurement/Controls/ItemDisplay.xaml.cs
@@ -51,6 +51,9 @@ namespace Procurement.Controls
                 itemImage.ContextMenu = getContextMenu();
                 contexted = true;
             }
+
+            //Reset the export to pob button incase user needs to export again
+            (buyoutControl.DataContext as SetBuyoutViewModel).IsDataInClipboard = false;
         }
 
         public static void ClosePopups()
@@ -183,6 +186,8 @@ namespace Procurement.Controls
             };
         }
 
+        private SetBuyoutView buyoutControl;
+
         private ContextMenu getContextMenu()
         {
             var vm = DataContext as ItemDisplayViewModel;
@@ -196,7 +201,7 @@ namespace Procurement.Controls
             var setBuyout = new MenuItem();
             setBuyout.StaysOpenOnClick = true;
 
-            var buyoutControl = new SetBuyoutView(item);
+            buyoutControl = new SetBuyoutView(item);
 
             if (Settings.Buyouts.ContainsKey(item.Id))
             {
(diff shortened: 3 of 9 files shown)


Procurement/ViewModel/SetBuyoutViewModel.cs
```diff
@@ -69,6 +77,30 @@ public void SetBuyoutInfo(ItemTradeInfo info)
...
+
+        private void GetPobData(object o)
+        {
+            try
+            {
+                Clipboard.SetText(item.PobData);
+
+                //Todo: Inject some sort of display framework - we don't want to spin up 
+                //a message box here...
+                MessageBox.Show("Path of building data copied to the clipboard", "Copied", MessageBoxButton.OK, MessageBoxImage.Information);
```
Stickymaddness: I'd remove the message box, it's going to be more annoying having it pop up than the value of seeing it will provide. We could maybe add something fancy like a border flash for confirmation, but in the meantime I'd go without the message box.

Stickymaddness:
This is an awesome addition!

Procurement/ViewModel/SetBuyoutViewModel.cs
aydjay: Fair comment, I'll work some feedback out inside SetBuyoutViewModel

aydjay:
I fucked up the rebase again...

I've googled and tried to understand where i'm going wrong but I don't understand!

```
git checkout master

git fetch upstream

git rebase upstream/master

git push 
```

Everything looks fine on the master branch (Upstream is stickymadness/procurement)

```
git checkout pobExport

git rebase origin/master

git push (This gives me a failure) 

hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

```

I read the hint about fast forwards and then do:

```
git pull --rebase
```
which results in this wierd history....

Please let me know where i'm going wrong!

thailyn:
Regarding the commits, I'm looking at the commit history, and I'm assuming the commits for this PR should be (in order, oldest to most recent):

```
6c359d6 - Add ability to export items into pob compatible format (4 days ago) aydjay
e1c15fc - Add mechanism to "Invert" a bool for visibility purposes (3 hours ago) aydjay
698b03e - (aydjay/pobExport) Modify clipboard data population feedback mechanism (3 hours ago) aydjay
```

If that isn't the case, tweak the commands below to compensate, or let me know and I'll revise them.

Whenever I rebase, for years, it is either with the `--onto` flag or `-i` (interactive).  I actually couldn't tell you what a plain `git rebase` does off the top of my head.  A `git rebase -i A` allows you to reorder or otherwise edit the commits in the half-open interval `(A, HEAD]`.  A `git rebase --onto A B C` moves the commits in the half-open interval `(B, C]` and applies them on top of `A`.

If I were fixing this, I would do the following:

``` bash
# Check out the tip of the branch we want to work with.
$ git checkout aydjay/pobExport
# Work off of a temporary branch, so we can still use the original if we mess up.
$ git checkout -b pob-temp
# Perform an interactive rebase so we can move the three commits we want to keep to the tip.
$ git rebase -i 161514b
```

Edit the file that comes up in your text editor so it looks like the following, then save and close it.  There could be additional commented lines, but leave them alone, since they're just instructions.  Note that the lines here are sorted oldest to newest.
``` text
pick 1e12739 Tidy up memory management of stash control
pick 3576c87 Update list of known divination cards
pick 8c1d2fb Procurement 1.25.0 Release
pick 23e40ac Remove email based login functionality
pick 6133c98 Fix imprint itemhover crash
pick 1748505 Fix item background bug by adding SetBackground() to AbstractStashTabControl
pick be17f82 Fix naming
pick 360a8bc Procurement 1.25.1 Release
pick 6c359d6 Add ability to export items into pob compatible format
pick e1c15fc Add mechanism to "Invert" a bool for visibility purposes
pick 698b03e Modify clipboard data population feedback mechanism
```

After the rebase finishes (no conflicts for me):
```bash
# Move the three commits at the tip of the now-interactivley-rebased branch onto the git tag "1.25.1".
$ git rebase --onto 1.25.1 50d7e57 pob-temp

# Move your local branch to point to the current commit. (Or do whatever other branch renaming you want.  The HEAD at this point is what you want to update this PR to be.)
$ git branch -f pobExport HEAD

# Force push the branch, since it won't apply cleanly.
$ git push -f ....
```

Alternatively to all of the above, you could check out `1.25.1` and cherry-pick those three commits, in order.  But that feels like cheating.

``` bash
$ git checkout 1.25.1
$ git checkout -b pob-temp-2
$ git cherry-pick 6c359d6 e1c15fc 698b03e 
# Do whatever branch moving/renaming/force pushing now.  The HEAD here should be the same as after the rebase --onto from the previous method.
$ git diff pob-temp pob-temp-2
# No differences printed.
```

Stickymaddness:
Haha oh man, git trial by fire there @aydjay Don't worry, it happens to everyone starting out 😄 

I can recommend [this video series](https://www.pluralsight.com/courses/mastering-git) to get a clearer understanding of how git works. 

@thailyn has given a good rundown of how to get things back on track, but in general if you need to update your branch after another PR has been merged I would just:

```
$ git pull --rebase
$ git push -f
```

aydjay:
@Stickymaddness Tell me about it! Half the reason I started contributing to procurement was to learn Git! 

@thailyn 

Thanks very much dude, here is exactly what I did.

```
git checkout 1.25.1
git checkout -b pobTemp
git cherry-pick 6c359d6 e1c15fc 698b03e
git push --set-upstream origin pobExport
git push origin HEAD:pobExport
```

POEApi.Model/Gear.cs
```diff
@@ -67,5 +70,51 @@ public int NumberOfSockets()
...
+                    var pobData = new StringBuilder();
+                    pobData.AppendLine(Name);
+                    pobData.AppendLine(TypeLine);
+                    pobData.AppendLine($"Unique ID: {Id}");
+                    pobData.AppendLine($"Item Level: {ItemLevel}");
+                    pobData.AppendLine($"Quality: {Quality}");
+
+                    if (Sockets != null && Sockets.Any())
+                    {
+                        var socketData = string.Join("=", Sockets.Select(x => x.ToPobFormat()));
```
thailyn: Add a comment explaining the format required by POB.  Are there any distinctions between links that are present and those that are missing?

Procurement/Controls/ItemDisplay.xaml.cs
```diff
@@ -183,6 +186,8 @@ private void BindSocketPopup(ItemDisplayViewModel vm)
             };
         }
 
+        private SetBuyoutView buyoutControl;
```
Stickymaddness: Why does this need to be a field?

README.md
```diff
@@ -12,7 +12,8 @@ Procurement is a Windows application for the online game Path Of Exile, that log
 * Search for gear based on item stats
 * Show where items are in your stash, that form part of vendor recipes
 * View your characters/gear
-* Oneclick generate forum trade posts that you can copy&paste into a trade post, which will have all your gear linked, verified and categorised in spoiler tags.
+* Oneclick generate forum trade posts that you can copy&paste into a trade post, which will have all your gear linked, verified and categorised in spoiler tags
+* Generate ![Path of Building](https://github.com/openarl/pathofbuilding) compatible item definitions
```
thailyn: Should have a hyphen at least between "Building" and "compatible".

Procurement/ViewModel/SetBuyoutViewModel.cs
```diff
@@ -69,6 +79,38 @@ public void SetBuyoutInfo(ItemTradeInfo info)
...
+        {
+            try
+            {
+                Clipboard.SetText(item.PobData);
+
+                IsDataInClipboard = true;
+            }
+            catch (Exception e)
+            {
+                MessageBox.Show("Unable to generate POB item data, please check log.","Error", MessageBoxButton.OK, MessageBoxImage.Error);
```
thailyn: Add a space between the first and second arguments.  Use a semicolon instead of a comma in the text to avoid a comma splice error.

Procurement/Controls/SetBuyoutView.xaml
```diff
@@ -52,8 +51,14 @@
 
 
         <TextBlock Grid.Row="3" Foreground="#FFAB9066" Text="Notes:" />
+
+        <TextBlock Grid.Row="3" Foreground="#FFAB9066" Text="POB Data copied to Clipboard" Grid.Column="1" Grid.ColumnSpan="2" Visibility="{Binding IsDataInClipboard, Converter={StaticResource bc}}"/>
+
+        <Button Content="Export to POB" ToolTip="Copy item data to clipboard in Path of Building format" 
+                Grid.Column="2" Height="25" Command="{Binding PobDataGenerationCommand}"  Grid.Row="3" Visibility="{Binding IsDataInClipboard, Converter={StaticResource bc}, ConverterParameter=Invert}"
```
thailyn: I don't like having the button disappear after the user presses it.  It doesn't appear clear to the user why it disappears, and when it will be back.  It doesn't hurt anyone for the user to just continue pressing the button, so we shouldn't limit him from doing so.

POEApi.Model/Socket.cs
```diff
@@ -10,5 +10,27 @@ internal Socket(JSONProxy.Socket s)
...
+        /// <summary>
+        /// Convert the Socket attribute to something consumable by path of building
+        /// </summary>
+        /// <returns></returns>
+        public string ToPobFormat()
+        {
+            switch (Attribute)
+            {
+                case "S":
+                    return "R";
```
thailyn: I'd rather see constants defined with these values instead of having magic strings.

Procurement/Controls/ItemDisplay.xaml.cs
aydjay: getContextMenu will assign it during ItemDisplay_MouseRightButtonUp.

Once the popup has been generated, I use this private field to cast the DataContext to get the ViewModel to assign IsDataInClipboard back to false so that the subsequent displays of the popup which have had the data copied to the clipboard will show the button for subsequent exports.

See:

https://github.com/Stickymaddness/Procurement/blob/1b950ec148e9e913e9fa0716c645d20d2ff4f1d3/Procurement/Controls/ItemDisplay.xaml.cs#L56

POEApi.Model/Gear.cs
```diff
@@ -67,5 +70,51 @@ public int NumberOfSockets()
...
+                    }
+
+                    if (CraftedMods != null && CraftedMods.Any())
+                    {
+                        CraftedMods.ForEach(x => pobData.AppendLine($"{{crafted}}{x}"));
+                    }
+
+                    return pobData.ToString();
+                }
+                catch (Exception e)
```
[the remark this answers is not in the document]
aydjay: In theory, no exceptions should happen but I can't predict which ones will occur so I'm just catching a generic one.

README.md
aydjay: I'm not sure what you mean exactly. This looks ok to me.
image

Procurement/Controls/SetBuyoutView.xaml
aydjay: The reason I hid the button is that there is limited space in the view and I didn't want to make too many changes here.

POEApi.Model/Gear.cs
aydjay: Great spot - I hadn't realised that slots which aren't linked are being linked up. *Fixed*

aydjay:
Ready when you are, thanks for the feedback 👍

README.md
thailyn: The phrase "Path of Building compatible" is a [compound adjective](https://www.quickanddirtytips.com/education/grammar/how-to-use-a-hyphen).  I was suggesting to change it to "Path of Building-compatible".  According to [some style guides](https://www.btb.termiumplus.gc.ca/tcdnstyl-chap?lang=eng&lettr=chapsect2&info0=2.04), though, when the compound adjective is made up of a noun and an adjective, the hyphen is only necessary when the phrase is otherwise ambiguous, which it isn't here.

POEApi.Model/Gear.cs
```diff
@@ -86,7 +86,21 @@ public override string PobData
 
                     if (Sockets != null && Sockets.Any())
                     {
-                        var socketData = string.Join("=", Sockets.Select(x => x.ToPobFormat()));
+                        //Links are denoted by "Groups"
```
thailyn: Add periods to the end of the comments (since you started them with capital letters).

POEApi.Model/Gear.cs
```diff
@@ -67,5 +70,65 @@ public int NumberOfSockets()
                 return components;
             }
         }
+
+        public override string PobData
```
thailyn: Move this into its own class (even if it's just a static method) and add tests to make sure we convert various `Gear` to strings correctly.

aydjay: It's not a method.

thailyn: You can change it into one.  I meant if the resulting class is just a static method.  Doing so is intended to make each class responsible for fewer things (a thing here being "convert a `Gear` to a PoB-compatible string") and make it easier to test.

aydjay: The way it's structure currently is that all items have the ability to provide a POB compatible string.

See:
https://github.com/Stickymaddness/Procurement/blob/ee8c9b0d483a05b012d26038dde74b944bf18842/POEApi.Model/Item.cs#L189

I can't envisage anything which isn't "Gear" being compatible for import into POB but I structured it this way so that other objects can provide their own data.

PR #955, opened 2019-02-09
Remove email based login functionality

Stickymaddness:
@aydjay @thailyn

### DIFF
--- a/POEApi.Model/POEModel.cs
+++ b/POEApi.Model/POEModel.cs
@@ -39,7 +39,7 @@ namespace POEApi.Model
             downOnlyMyCharacters = bool.Parse(Settings.UserSettings["DownloadOnlyMyCharacters"]);
         }
 
-        public bool Authenticate(string email, SecureString password, bool offline, bool useSessionID)
+        public bool Authenticate(string email, SecureString password, bool offline)
         {
             if (Transport != null)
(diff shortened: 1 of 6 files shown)


thailyn:
I'm hesitant to simply gut all the non-sessionID functionality, since I think it can be repaired, given some more effort.  However, we can always look at the revision history and find the old code.

If you do keep it as-is, perhaps add a TODO to either fix username/password login or determine it is indeed impossible, and reference this PR?

Stickymaddness:
Honestly I'm tempted to just remove it completely, since I think sessionid is the most widely used as the majority of 3rd party tools tend to use SessionID these days.

Alternatively if you'd like to take a stab at repairing the email functionality, I could reduce the changes in this PR to simply have the xaml changes as it's confusing to users having the email/password login when it doesn't work. That way the UI at least represents that you can only currently login with sessionID but it would be a small change to revert if email is working again.

thailyn:
I am interested in trying to try to repair it -- as that is how I logged in previously -- but it isn't that high a priority since SessionID does work.  But, yes, we are definitely overdue in updating the UI so users know only SessionID works right now.  Either way you want to make the change is fine to me, since we do have the revision history.

Stickymaddness:
Alright, in that case I'm going to merge as is. I'll happily merged the functionality back in when you do take a stab at getting email to work again.

PR #956, opened 2019-02-09
Fix imprint itemhover crash

Stickymaddness:
I remember running into items with the "<<set:MS>><<set:M>><<set:S>>" text in the past, from what I remember we used a quick and dirty string replace initially, but found a cleaner fix which I can't seem to find. 

Any ideas @aydjay @thailyn ?

### DIFF
--- a/POEApi.Model/Property.cs
+++ b/POEApi.Model/Property.cs
@@ -12,16 +12,23 @@ namespace POEApi.Model
 
         internal Property(JSONProxy.Property property)
         {
-            this.Name = property.Name;
+            Name = property.Name;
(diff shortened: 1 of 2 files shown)


thailyn:
Do you remember what items this was with?  My only concern is if we swallow the text completely we won't be able to notice when it happens and possibly fix it.

Stickymaddness:
I ran into it while fixing the imprint item hover crash, but it has popped up on other items in the past - I just can't recall where.

My thinking with this change is to catch any items that might show up in the future with the prepended text, and strip it out.

Issue #949, opened 2019-02-05
Item quantity/rarity search not working


nqrcqn:
When I choose either Item Quantity or Item Rarity on Advanced Search no items show up even though I have quite a few of them.

Edit: Increased Attack Speed and Increased Cast Speed return no results as well.

nqrcqn:
At the very least,

Increased Cold Damage
Increased Fire Damage
Increased Lightning Damage

have the same issue, from the looks of it, they aren't being checked in correct case like "**i**ncreased Cold Damage"

nqrcqn:
Off topic, if anyone is up for it, more phrases can be supported like:

15% increased Item Quantity per White Socket
Nearby Allies have 30% increased Item Rarity
to Cold/Fire/Lightning and Chaos Resistances

aydjay:
![image](https://user-images.githubusercontent.com/9460838/52285269-f45ddc00-295d-11e9-877d-9c1cc0800c25.png)

Looks ok for me, this is using V1.24.2

If you could link the items that have the mods that you want filters included for and the appropriate bin files, that have those items that would help alot.

The bin files are in a folder with your account name next to where you have procurement installed.

nqrcqn:
Looks like an issue with handling OS locale, lower case of I or upper case of i both differs from the english equivalent on my system.

aydjay:
What locale is your system set too?

nqrcqn:
Turkish

PR #962, opened 2019-02-15
Fix item search for certain system locales

nqrcqn:
As mentioned in #949 using upper case "I" is causing issues on some system locales, this should fix it. Uses same lower case "i" as it is used on items.

### DIFF
--- a/Procurement/ViewModel/Filters/ForumExport/AttackSpeed.cs
+++ b/Procurement/ViewModel/Filters/ForumExport/AttackSpeed.cs
@@ -8,7 +8,7 @@
         }
 
         public AttackSpeed()
-            : base("Increased Attack Speed", "Increased Attack Speed", "Increased Attack Speed")
+            : base("Increased Attack Speed", "Increased Attack Speed", "increased Attack Speed")
         { }
     }
 }
--- a/Procurement/ViewModel/Filters/ForumExport/CastSpeed.cs
+++ b/Procurement/ViewModel/Filters/ForumExport/CastSpeed.cs
@@ -8,7 +8,7 @@
         }
 
         public CastSpeed()
-            : base("Increased Cast Speed", "Increased Cast Speed", "Increased Cast Speed")
+            : base("Increased Cast Speed", "Increased Cast Speed", "increased Cast Speed")
         { }
     }
 }
--- a/Procurement/ViewModel/Filters/ForumExport/IncreasedDamageFilter.cs
+++ b/Procurement/ViewModel/Filters/ForumExport/IncreasedDamageFilter.cs
@@ -15,21 +15,21 @@
     internal class IncreasedDamageFilterCold : IncreasedDamageFilter
     {
         public IncreasedDamageFilterCold()
-            : base("Increased Cold Damage", "Increased Cold Damage", "Increased Cold Damage")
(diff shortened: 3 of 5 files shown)


thailyn:
If all you're doing is changing the case of a few strings to find, wouldn't it be better to change the `Regex` in `Procurement/Procurement/ViewModel/Filters/ForumExport/StatFilter.cs` to ignore case? (See the first code example [here](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex?view=netframework-4.7.2).)  Or would that not solve this problem with locales here?

nqrcqn:
Just tested the build and it fixed the issue for me. It may not be the ideal fix and there may be more strings that check for upper case I, but those 7 searches now work correctly on my system.

nqrcqn:
What you mentioned could solve it, I just tried to fix it with my limited knowledge

aydjay:
I spent a hour or so looking into this issue when you reported it. I focused on the entry from the dialog.

I found this guideline from ms that explained the problem in detail.

https://docs.microsoft.com/en-us/globalization/locale/sorting-and-string-comparison

The fix you have proposed would resolve it for these recipes but wouldn't fix it for all recipies going forward. Investigate what @thailyn has suggested coupled with the information above and we might be able to solve it for all cases.

nqrcqn:
I don't think a language specific solution is required in this case, unless you know how to do it, and easily.

For example, "Intelligence" is also checked with upper case "I", but since it matches the case in item text, it doesn't cause the same issue, so cleanest solution is still using the same case as it appears on items, and so far I haven't encountered this issue on other searches.

aydjay:
A language agnostic solution would be best. I think all the changes you have proposed can ve reverted and instead change: 

https://github.com/aydjay/Procurement/blob/43355d98df199ab7d9cd13c57df3e761770b1e95/Procurement/ViewModel/Filters/ForumExport/StatFilterBase.cs#L17

to: 

this.stats = stats.Select(stat => new Regex(stat.ToLowerInvariant(), RegexOptions.Singleline | RegexOptions.IgnoreCase)).ToList();

This would mean it doesn't matter what the casing provided to the filter constructor eg: ItemQuantity and ItemRarity.

@thailyn What do you think?

The other area that suffers is when the Turkish lower case dotless i/ı is used to filter the list of available filters in the UI as described in #949

nqrcqn:
If you provide a build, I'd test and report back the results

aydjay:
I don't have time to investigate this fully at the moment, I just thought I would give you some pointers to get investigating with. You certainly seem to be on the right track.

nqrcqn:
Your fix works, I have tested with this build: https://ci.appveyor.com/project/nqrcqn/procurement/builds/22483773

But my commit history is long, if you open a pull request with that small change it would be better

aydjay:
Great! 

Don't be intimidated by Git, you can do it! 

It looks like you've been making changes against your master branch. As far as I'm aware this isn't best practice. Have a little look at this, you will want to adjust your workflow accordingly

http://rogerdudler.github.io/git-guide/

You will want to reset your local changes and then make a new branch upon which you want to apply your fix too.

Once you've made the new branch close this pull request and open a new one against the new branch. 

You may also want to look into the filtering of the available filters while you are at it.

nqrcqn:
new pull request #964

PR #964, opened 2019-02-19
Fix item search not working on some systems

### LINKED ISSUE #949: Item quantity/rarity search not working

nqrcqn:
Fixes #949


aydjay:
Looks good to me.

Stickymaddness:
Nice work 👍

PR #968, opened 2019-02-27
Fix currency image rendering

### LINKED ISSUE #966: Currency image incorrect rendering
With the latest build (As of today)

image

- [x] All the currency images don't line up
- [x] The inbuilt stack that used to be on the image (IIRC) is gone

Old image looked like this:
https://web.poecdn.com//image//Art//2DItems//Currency//CurrencyAddModToMagic.png?scale=1&scaleIndex=3&stackSize=37&w=1&h=1&v=97e63b85807f2419f4208482fd0b4859

Notice the stack and scale in the URL, new response from GGG is as follows.

https://web.poecdn.com//image//Art//2DItems//Currency//CurrencyAddModToMagic.png?scale=1&w=1&h=1&v=97e63b85807f2419f4208482fd0b4859

We could add the stack and scale into the URL by hand or we could remake the display. Not sure what is the best option.

aydjay:
Closes #966 

Thanks, @Stickymaddness for your investigation - I just made a few tweaks to the xaml as you suggested.

### DIFF
--- a/Procurement/Controls/ItemDisplay.xaml.cs
+++ b/Procurement/Controls/ItemDisplay.xaml.cs
@@ -23,6 +23,24 @@ namespace Procurement.Controls
         private TextBlock textblock;
         private ItemDisplayViewModel _viewModel;
 
+        public static readonly DependencyProperty HeightScaleProperty = DependencyProperty.Register(
+            "HeightScale", typeof(double?), typeof(ItemDisplay), new PropertyMetadata(null));
+
+        public double? HeightScale
+        {
+            get { return (double?) GetValue(HeightScaleProperty); }
+            set { SetValue(HeightScaleProperty, value); }
+        }
+
+        public static readonly DependencyProperty WidthScaleProperty = DependencyProperty.Register(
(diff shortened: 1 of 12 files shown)


aydjay:
Tagging @nqrcqn, maybe you can test it out from the build artifacts here: https://ci.appveyor.com/project/Stickymaddness/procurement/builds/22692791/artifacts

Procurement/Controls/ItemDisplay.xaml.cs
```diff
@@ -68,9 +86,16 @@ private void ItemDisplay_Loaded(object sender, RoutedEventArgs e)
             {
                 var i = vm.getImage();
                 itemImage = i;
-
+                
                 if (i != null)
                 {
+                    //See: https://github.com/Stickymaddness/Procurement/issues/966
```
thailyn: Include a summary here of the situation, as well as the reference. (The link might not always be valid.)

Stickymaddness:
Nice work Alex, great that we're able to change this in the view layer!

PR #969, opened 2019-03-01
Fix naming... a long time coming!

Stickymaddness:
It's done


aydjay:
Yay, c# naming conventions.

thailyn:
I really need to get around to adding some StyleCop rules to make these naming conventions "official" and can enforce new code following them.

PR #971, opened 2019-03-01
Fix crafting slot positioning in currency tab

Stickymaddness:
This fixes the placement of items in the crafting slot by setting the scale on the item style. The caveat with this is that because the scaling is on the xaml, it does not apply to the sockets and links, thus the link images overlap with some of the sockets.

#966


aydjay:
I tried messing about with the scaling on this slot myself but to no avail. I've pulled the artifact back from this PR but it's still displaying the same behaviour as previously shown.

image

Interestingly the currency breakdown seems borked now 😕 

[d347hm4n.zip](https://github.com/Stickymaddness/Procurement/files/2920677/d347hm4n.zip)

This is my stash incase you are interested in checking out the behaviour yourself.

Stickymaddness:
Realised the problem is that I tested with a 2w3h item (body armor) but not only body armor goes into that slot! 

So issue is that the Height/WidthScale property is going to be different, depending on the item that's in the slot.

Stickymaddness:
I seem to have broken Currency breakdown in #969

aydjay:
A converter bound against the item w and h should do the trick then, I think.

https://github.com/Stickymaddness/Procurement/blob/480ae354a615ba60d8611805fac66397a6ef26cd/POEApi.Model/Item.cs#L34

Stickymaddness:
Nice thinking, even if the ratios are hardcoded for now.

Being able to scale item + sockets/links from a single scale property would be the correct way to fix this for the display of all items. I could see this fixing quad tab as well.

Stickymaddness:
Added a converter to handle different item types, it still needs a bit of tweaking to get vertical placement perfect, but it's at least aligned properly in the slot.

I've also added backgrounds to slotted items, which was missing:

Stickymaddness:
Haha, not at all. I used a macro in vim to generate most of the changes.

Can you give me an example of what the xaml would look like for this?

aydjay:
https://github.com/Stickymaddness/Procurement/blob/7c52d06cee949cbcb9a09f6da7b738f6d55f1c8d/Procurement/Controls/ItemDisplay.xaml#L13

Off the top of my head

https://www.wpf-tutorial.com/styles/trigger-datatrigger-event-trigger/

You'll want a data trigger. Against the identified property on the item which you will expose via a property on the view model.

Stickymaddness:
Having looked into this a bit, the background being on the ItemDisplay is how it should have been from the start - for all items. This would also fix certain item types, such as essences, that currently do not have their background correctly displayed. 

However this does have a few gotchas that will make the change bigger than I would like it to be right now, such as Shaper/Elder items that need their background on top of the blue/red background which would necessitate moving that code to ItemDisplay (which is probably a better place for it). Secondly, items in essence tabs do _not_ have a blue background, which means that we would need a way to set that, presumably a property on ItemDisplay - this would also be needed by the crafting slot in currency tabs, as the slot is larger than the item so if the background is restricted to the item only part of the slot will have the blue background instead of the whole slot.

With Synthesis launching in less than a week, I'm going to merge this as so that it can be released sooner rather than later considering the high impact for users.

Moving the background to ItemDisplay as you suggested is **definitely** the way this should be handled in the future, removing the logic from the tabs will simplify things a lot.

aydjay:
Cool, looks like we need to raise a couple of issues and tag them up appropriately so we can quantify the outstanding items. I'll read through the backlog and the recent pr's and see what we get.

PR #972, opened 2019-03-01
Fix stacksize not showing in special tabs for all items

Stickymaddness:
The intent for this change was to fix stack amount not showing for all items in special tabs, for example sacrifice fragments in fragment tabs, but has resulted in StackableItem being removed. 

/cc @aydjay


Procurement/ViewModel/ItemDisplayViewModel.cs
```diff
@@ -57,7 +57,7 @@ public SolidColorBrush ItemFilterBrush
                                         : new SolidColorBrush(Colors.Transparent); }
         }
 
-        public bool IsStackSizeVisible => Item is StackableItem;
+        public bool IsStackSizeVisible => Item?.StackSize > 1;
```
aydjay: You won't see it if your stack size is one, which is a valid amount of stacks.

Stickymaddness: Bah, nice catch I'll fix it.

PR #974, opened 2019-03-02
 Add visual support for Elder and Shaper items 

Stickymaddness:


POEApi.Model/Item.cs
```diff
@@ -111,8 +115,11 @@ protected Item(JSONProxy.Item item)
 
             TradeX = X;
             TradeY = Y;
-            TradeInventoryId =InventoryId;
+            TradeInventoryId = InventoryId;
             Character = string.Empty;
+
+            if (item.Elder || item.Shaper)
+                BackgroundUrl = ItemBackgroundUrlBuilder.GetUrl(this);
```
aydjay: I'm not personally a massive fan of skipping braces on single if statements, you're the owner though so it's entirely up to you.

Procurement/Controls/AbstractStashTabControl.cs
```diff
@@ -106,12 +106,34 @@ public void Control_Loaded(object sender, RoutedEventArgs e)
 
         protected void SetBackground(Grid childGrid, Item item)
         {
-            if (item is Gear && (item as Gear).Rarity != Rarity.Normal && (item as Gear).Explicitmods == null)
+            if (item is Gear == false)
```
aydjay: var gear = item as Gear
if(gear == null)
    return;

save's boxing/unboxing it twice.

Procurement/Controls/AbstractStashTabControl.cs
```diff
@@ -106,12 +106,34 @@ public void Control_Loaded(object sender, RoutedEventArgs e)
...
+        }
+
+        private static void SetItemBackground(Grid childGrid, Item item)
+        {
+            if (!item.HasBackground)
+                return;
+
+            try
+            {
+                (childGrid.Children[0] as ItemDisplay).Background = new ImageBrush(ApplicationState.BitmapCache[item.BackgroundUrl]);
```
aydjay: Child grid may not have children. Check before accessing. (Unlikely I appreciate)

aydjay: To top this all off, might be an idea to pull all the appropriate images that we might need and bundle them with the application. Will save going across the wire to download the images. The you can pull the appropriate resource out of the dictionary. 

https://docs.microsoft.com/en-us/dotnet/api/system.windows.application.resources?view=netframework-4.7.2

Just a suggestion of course, doing it like this is a perfectly pertinent approach.

POEApi.Model/Item.cs
Stickymaddness: We have both styles all over Procurement, I should probably decide on one and update it across the code base as I did with the renaming.

Procurement/Controls/AbstractStashTabControl.cs
Stickymaddness: I actually considered that starting out, but think of it this way - there are 144 slots in a standard tab, so that's a 144 images to cover 1h1w shaper items. Then think of the number of different item sizes, and the number of spaces they can each occupy in a stash - again it's a unique image for each.

When GGG inevitably changes the background it will all have to be pulled down again, and if/when new item types are released in the future with an animated background the number of images will double.

It saves time down the wire (once), but increases the size of the Procurement zip for every release - whether you have those items or not.

Procurement/Controls/AbstractStashTabControl.cs
Stickymaddness: I tend to default to 'is' to avoid unnecessary casting, but in this case it should always be gear so saving the boxing/unboxes makes sense as it will have to cast right after anyway.

Procurement/Controls/AbstractStashTabControl.cs
aydjay: Fair shout, this also innoculates the code from any more of GGG's changes

PR #975, opened 2019-03-02
Update tab dropdown to use tab colours

Stickymaddness:


Stickymaddness:
That'd be a nice improvement, would be great to have it styled as per the official one as well

PR #978, opened 2019-03-06
Add a setting to use short recipe descriptions.

thailyn:
If this setting, "UseShortRecipeDisplayDescriptions", is false, the recipe descriptions are displayed as before.  If it is true, shorter (but more cryptic) descriptions are used, which allow the number of matches to be visible.  This currently only applies to the SameBaseTypeRecipe descriptions, but could be added to other long recipe descriptions.

image


Procurement/ViewModel/Recipes/Recipe.cs
```diff
@@ -38,5 +39,13 @@ public virtual IEnumerable<RecipeResult> Matches(IDictionary<Tab, List<Item>> it
             }
             return Matches(flatItems);
         }
+
+        protected virtual bool GetShouldUseShortRecipeDescriptions()
```
aydjay: Doesn't need to be virtual - it's not overridden.

Procurement/ViewModel/Recipes/SameBaseTypeRecipe.cs
```diff
@@ -34,30 +34,47 @@ public override string Name
 
         public override string GetResultName(RecipeResult result)
```
aydjay: We should have a constructor parameter that takes a "Long name" and a "Short name" for every recipe otherwise people will forget to provide on when a recipe is implemented. This should be applied to all existing recipes.

thailyn: Honestly, I'm only making this PR so I don't have to keep rebasing a local change to shorten these recipe names for myself.  Also, especially since each recipe can have arbitrarily many names/descriptions, we would need to take lists of short and long names in the constructor (or something more complicated to ensure the long and short names match up to each other).  Moreover, the short names are obfuscated, so I would rather not encourage using them unless we have to (that is, unless they're too long).

Procurement/ViewModel/Recipes/Recipe.cs
thailyn: Done.  I usually err on the side of making more stuff virtual so it's easier down the road, but, indeed, it isn't being overridden right now.

Procurement/ViewModel/Recipes/SameBaseTypeRecipe.cs
aydjay: That's a fair comment for sure, I ran I to the long name issue with the shaper elder set recipe's. Similar problem. Maybe reworking the UI somehow may be a longer term solution.

thailyn: Ah, right, the elder/shaper sets have long names.  I'll add short names for them.

I agree that a UI rework is probably a better longer term solution.  I just don't want to figure out how to do any of that right now. :persevere:

thailyn: Alright, I added short names for each of the `RareSetRecipe` instances.

Procurement/ViewModel/Recipes/Recipe.cs
```diff
@@ -38,5 +39,13 @@ public virtual IEnumerable<RecipeResult> Matches(IDictionary<Tab, List<Item>> it
             }
             return Matches(flatItems);
         }
+
+        protected bool GetShouldUseShortRecipeDescriptions()
+        {
+            bool result = false;
+            bool.TryParse(Settings.UserSettings.GetEntry("UseShortRecipeDisplayDescriptions"), out result);
```
Stickymaddness: Beyond the scope of this PR, but settings are in dire need of a rework. Needing to stuff like this is just horrible.

PR #984, opened 2019-03-11
Cloudflare lib as submodule

thailyn:
Change to use the `CloudFlareUtilities` library as a submodule, instead of a NuGet package, so we can use unreleased versions.  There is both a merge pull request and a pending pull request which improve the reliability of authenticating against CloudFlare.  With both, it appears to be completely reliable!

Two caveats:
* The `CloudFlareUtilities` solution uses Visual Studio version 15 (2017), so it appears we need to use that version now, as well (I was getting issues with the compiler version).
* The revision used in the `CloudFlareUtilities` is for a pending pull request.  I think this is a very ugly way of going about this, but the unreliable nature of the CloudFlare authentication is probably Procurement's biggest issue right now (and other tools don't seem to have it).  I think it's worth jumping through a few hoops to get this working better.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/CloudFlareUtilities"]
+	path = lib/CloudFlareUtilities
+	url = https://github.com/elcattivo/CloudFlareUtilities.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "lib/CloudFlareUtilities"]
 	path = lib/CloudFlareUtilities
-	url = https://github.com/elcattivo/CloudFlareUtilities.git
+	url = git@github.com:Eric898989/CloudFlareUtilities.git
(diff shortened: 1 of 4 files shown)


thailyn:
I think how AppVeyor works needs to be updated to work with submodules.  Partial explanation [here](https://www.appveyor.com/docs/how-to/private-git-sub-modules/#the-problem-with-private-sub-modules).

I think adding a command like `git submodule update --init --recursive` would work.

aydjay:
I don't think we should deviate from the Master branch of cloudflareutils for these reasons:

1. There is no gaurantee that "Eric898989" is going to continue maintaining his fork
2. Increased overhead on us to keep an eye on the PRs against cloudflareutils
3. Although it's not perfect, it does work the majority of the time

thailyn:
Well, there's no guarantee the parent CloudFlareUtiltiies repo will continue to be maintained, either. ^_^ I could make a fork of the repo with the pull request and use that, if we want to make sure that at least sticks around.

Personally, since the CloudFlare authentication is so troublesome (it usually *doesn't* work for me, and it's the cause of most issues we get), I think it's worth it to go through this extra effort.

PR #993, opened 2019-03-13
Allow gems that are socketed in items to appear in the Gem Distribution

### LINKED ISSUE #991: Portal gem not listed under Gems tab
The `Portal` gem is not listed under the Gems tab.

bug_portal_gem_not_listed_under_gems

aydjay:
Closes #991


Stickymaddness:
I could swear I remember this already being added a few years back...

PR #994, opened 2019-03-13
Full delve item forum export support

### LINKED ISSUE #992: Scorched Fossil not added to forum export
I haven't had a chance to try with other fossils, but the 3 scorched fossils I have in my stash don't get added to forum export, I think it worked fine during Betrayal or a previous Procurement version. They used to show up under Currencies

aydjay:
Closes #992 
Add's full support for fossils and resonators.

I refactor the visitor generation code while I was there, there was alot of boilerplate, should be even simpler to generate Visitors from now on.


aydjay:
@nqrcqn Please test this, should allow you to sell fossils and resonators.

nqrcqn:
Yes, fixes fossils

aydjay:
Because this PR updates the hover for fossils to fix Scarabs and Splinters (and other items in the future) I'm defaulting the item hover header to the currency one.

Closes #998 too.

@nqrcqn Can you check it with the latest artifact from my latest commit. Thanks.

nqrcqn:
Can't comment on the defaulting but Scarabs have the same white/gray background as Offering/Divine Vessels, so I think those should be Rarity.Normal at least

nqrcqn:
I just realized splinters aren't added to forum export at all, so that explains having no currency background as well.

Procurement/ViewModel/ItemHoverViewModelFactory.cs
```diff
@@ -70,13 +70,11 @@ internal static ItemHoverViewModel Create(Item item)
...
 
-            if (item is Currency || item is Sextant || item is Essence || item is Fossil)
-                return new CurrencyItemHoverViewModel(item);
-
             if (item is Prophecy)
                 return new ProphecyItemHoverViewModel(item);
 
-            return new ItemHoverViewModel(item);
+            //Currency, sextant, essence, fossil, resonator, scarabs and splinters
+            return new CurrencyItemHoverViewModel(item);
```
Stickymaddness: Are we sure it's safe to default to currency hover? What about if new types get added in the future?

aydjay: Your call, but the if to get to CurrencyItemHoverViewModel was getting out of hand. 7 items and no doubt more would be on the way.

aydjay:
Bumping this PR for a merge please @Stickymaddness

Procurement/ViewModel/ItemHoverViewModelFactory.cs
Stickymaddness: I'd rather do that and keep the order of precedence in tact. The class is probabaly due for a bit of a cleanup, as much as I hate them maybe a switch would be clearer with the number of conditionals, or separate hover view models for sextant/essence/fossil.

aydjay: I've restored it to a long IF statement as per your request.

nqrcqn:
I'm testing the latest build, one of the recent merges broke tab refreshing. Tabs do not refresh unless I close and restart Procurement.

aydjay:
@nqrcqn If you could release a detailed bug report in a new that would be amazing. 👍

nqrcqn:
The only detail is clicking Refresh All/Used Tabs does not refresh the tabs anymore.

PR #1005, opened 2019-03-21
Revert "Item Background generation refactor"

Stickymaddness:
Reverts Stickymaddness/Procurement#989

@aydjay Unfortunately I'm going to have to revert this, as it (visually) breaks the currency tab, as the existing item slot border backgrounds weren't removed which means orbs have double backgrounds and the crafting slot item has two shades of background. In addition because the essence tab items aren't properly aligned it's much more noticeable with the backgrounds.

### DIFF
--- a/Procurement/App.xaml
+++ b/Procurement/App.xaml
@@ -10,7 +10,7 @@
             <ResourceDictionary.MergedDictionaries>
                 <ResourceDictionary Source="/Procurement;component/Controls/TabControlStyle.xaml" />
             </ResourceDictionary.MergedDictionaries>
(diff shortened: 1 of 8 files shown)


aydjay:
This is because you have old data that has the old images, you need to pull down a fresh stash.

nqrcqn:
So is it safe to revert this revert?

aydjay:
I would have to check this pr out again and let you know

nqrcqn:
I think it would be nice to fix/revert this before the next league

aydjay:
For sure that would be great - we have too many outstanding PR's at the moment though. I won't have time to address this before the league drops on Friday. Be more important to get a quick release with the updated visual for blight league.

Issue #1013, opened 2019-04-05
Add PS4/XBOX support


Flenst:
The get-item and get-stash-items URL needs a simple "&realm=sony" or "&realm=xbox" added, maybe with a dropdown at Login. Rest works the same. My fork only supports PS4, adding dropdown and multi platform support is beyond my capabilities ;)

aydjay:
Do you have any api documentation that shows exactly what is needed to "Talk" to the PS4 realm?#

Adding in a mechanism as you have suggested with the UI isn't a massive piece of work but knowing exactly what has changed will help a lot.

Flenst:
No api documentation, I looked at my network traffic in browser to find out where is the difference to the PC. The realm names are in the .js of the main site. My simple commit https://github.com/Flenst/Procurement-PS4/commit/a512c1d0717bd9be0aa530506dad7798b70bb03c makes it fully functional with PS4 accounts. Using xbox instead of sony should make it fully functional with xbox accounts.

aydjay:
I'm a bit reluctant to implement some changes for "Undocumented" parts of the API, as there could be other messages that need to be modified for it to work correctly. I have noticed in your commit that you haven't applied the realm parameter to the other API methods: LoginURL, AccountNameURL and CharacterURL for example.

I don't own an XBox or PS4 either so testing will be tricky.

@Stickymaddness What's your opinion?

Flenst:
Yes, those gave me the correct results without a need to specify the realm. The realm was necessary for the stash tabs and inventory tabs, at launch they threw an error without the realm specified.

killa78:
@aydjay  I know that Openarl (Path of Building) has utilized the new API  to allow it to import from the username etc.  See [The Import Tab Code within PoB](https://github.com/Openarl/PathOfBuilding/blob/master/Classes/ImportTab.lua)

aydjay:
This can be closed

PR #1019, opened 2019-04-30
Console support

aydjay:
Provides the choice of which "Realm" to connect to.

image

Closes:

#1013 

Tagging: @Flenst @killa78 for them to test it as I don't own a console.

### DIFF
--- a/POEApi.Model/Events/AuthenticateEventArgs.cs
+++ b/POEApi.Model/Events/AuthenticateEventArgs.cs
@@ -3,10 +3,13 @@
     public class AuthenticateEventArgs : POEEventArgs
     {
         public string Email { get; private set; }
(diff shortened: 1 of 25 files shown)


Stickymaddness:
As far as I'm aware Procurement is working fine for console players without it passing the realm parameter, however if that's the direction GGG's API has gone I think it makes sense to conform to it because sooner or later it will probably break without it.

That being said, could we get a the PC Master Race mascot popping up if you select a platform other than PC? (I'm kidding)

Stickymaddness:
The merge build for this failed (I [fixed it](https://github.com/Stickymaddness/Procurement/commit/79f3307fb6c9fb64e43b0c0ab65f719732f59f48)), it seems the error wasn't caught by the PR build.

PR #1025, opened 2019-06-10
Legion Support

aydjay:
Allow Splinters and Emblems to be exported and filtered

- [x] Add Incubators, splinters (Legion and breach), breach stones and emblems to export
- [x] Add Incubation information into Item hover
- [x] Update UI for Legion


POEApi.Model/LegionType.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace POEApi.Model
+{
+    public enum LegionType
```
thailyn: Let's call this `LegionFaction`, so it's clear what is being referred to.

POEApi.Model/LegionType.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        EternalEmpire,
+        Karui,
+        Maraketh,
+        Templar,
+        Vaal, 
+
+        Unknown
```
thailyn: It's not generally done with enums in Procurement, but we should make `Unknown` the default value (by making it [have a value of 0](https://stackoverflow.com/a/4967673)).

POEApi.Model/ProxyMapper.cs
```diff
@@ -209,6 +209,18 @@ internal class ProxyMapper
             {"Uul-Netol", BreachType.UulNetol}
         };
 
+        private static readonly Dictionary<string, LegionType> _legionMap = new Dictionary<string, LegionType>
+        {
+            //Splinters are called Timeless Eternal Empire Splinter
```
thailyn: We should make these comments more explicit.  Something like: ```Eternal Empire splinters are called "Timeless Eternal Empire Splinter".``` Similar change for the emblems comment (the comment only applies to the Eternal Empire emblem).

POEApi.Model/ProxyMapper.cs
```diff
@@ -313,6 +325,21 @@ public static BreachType GetBreachType(JSONProxy.Item item)
...
+        public static LegionType GetLegionType(JSONProxy.Item item)
+        {
+            try
+            {
+                return _legionMap.First(m => item.TypeLine.Contains(m.Key)).Value;
+            }
+            catch (Exception ex)
+            {
+                Logger.Log(ex);
+                Logger.Log("ProxyMapper.GetBreachType Failed! ItemType = " + item.TypeLine);
```
thailyn: s/GetBreachType/GetLegionType

Procurement/ViewModel/Filters/ForumExport/LegionFilter.cs
```diff
@@ -0,0 +1,16 @@
+﻿using POEApi.Model;
```
thailyn: If this file has only `IncubatorFilter`, it should be named `IncubatorFilter.cs`.

Procurement/ViewModel/ItemHoverViewModel.cs
```diff
@@ -104,6 +118,18 @@ public ItemHoverViewModel(Item item)
                 ExperienceDenominator = gem.ExperienceDenominator;
             }
 
+            if (IsIncubatorProgressVisible)
+            {
+                IncubatorNumerator = item.IncubatedDetails.Progress;
+                IncubatorDenominator = item.IncubatedDetails.Total;
+                Incubating = $"Incubating {item.IncubatedDetails.Name}";
+                IncubationLevel =  $"Level {item.IncubatedDetails.Level}+ Monster Kills";
+                if (Item.IncubatedDetails.Total > 0 )
```
thailyn: No space before the `)`.

POEApi.Model/JSONProxy/Stash.cs
```diff
@@ -26,6 +26,14 @@ public class Requirement
         public int DisplayMode { get; set; }
     }
 
+    public class IncubatedDetails
```
thailyn: Why not name it `IncubatorDetails`?

POEApi.Model/JSONProxy/Stash.cs
```diff
@@ -26,6 +26,14 @@ public class Requirement
...
+    {
+        public string Name { get; set; }
```
thailyn: We should have comments explaining exactly what these values mean. (E.g., how `Progress` and `Total` relate isn't intuitively obvious.)

POEApi.Model/JSONProxy/Stash.cs
```diff
@@ -26,6 +26,14 @@ public class Requirement
...
+        public int Level { get; set; }
```
thailyn: Is `MinimumLevel` a more accurate name for all uses of the variable?

Procurement/ViewModel/ItemHoverViewModel.cs
```diff
@@ -104,6 +118,18 @@ public ItemHoverViewModel(Item item)
...
+                {
+                    IncubatorProgress = double.Parse(item.IncubatedDetails.Progress.ToString()) / double.Parse(item.IncubatedDetails.Total.ToString());
```
thailyn: We don't need to convert the values to strings first, do we?  We could simply convert one from an `int` to a `double` and get a `double` result.

POEApi.Model/LegionType.cs
aydjay: Done - rename additional symbols

POEApi.Model/ProxyMapper.cs
aydjay: I think it's clear enough.

POEApi.Model/ProxyMapper.cs
aydjay: Fixed, we should really leverage [CallerMemberName] in the logger.

Procurement/ViewModel/Filters/ForumExport/LegionFilter.cs
aydjay: Done

POEApi.Model/JSONProxy/Stash.cs
aydjay: You apply the Incubator to the item you have equipped. Thus making it the "Details" which have been "Incubated"

Procurement/ViewModel/ItemHoverViewModel.cs
aydjay: Fixed, we should get style cop for these sorts of nits.

POEApi.Model/JSONProxy/Stash.cs
aydjay: These are tied to the properties that come over the wire on the JSon objects.

POEApi.Model/JSONProxy/Stash.cs
aydjay: As above

Procurement/ViewModel/ItemHoverViewModel.cs
aydjay: Just learn about "Convert.ToDouble(anInt)" thanks to this comment.

Thanks

aydjay:
Applied the changes that @thailyn recommended

thailyn:
The text for "Incubating X" on the item tooltip should be the same color as master-crafted mods (though the color we're using for the latter isn't quite correct, either):
image

How it currently looks in Procurement:
image

Also, with the new login screen background, it's practically impossible to read the "Refresh on Login" text.  Can we add a solid background behind that text?
image

aydjay:
Thanks for the comments on the "Look" and "Feel"

I have made some very minor adjustments and I think it's improved. (I would though :P)

POEApi.Model/JSONProxy/Stash.cs
Stickymaddness: One of the big mistakes in Procurement was tying its internal models to the json ones that come from the API, they should be separated and mapped across. It would have saved us so many times in the past with so many issues.

Stickymaddness:
Nice work, especially on some of the extras like the background and login styling.

PR #1032, opened 2019-06-13
Restore ITransport GetStash with forceRefresh.

### LINKED ISSUE #1029: Refreshing tabs is broken
As I reported previously: https://github.com/Stickymaddness/Procurement/pull/994#issuecomment-490991339

Refreshing just doesn't work, and after the new release, Procurement now crashes after the failed refresh

1.26.1 still refreshes without an issue

[debug.txt](https://github.com/Stickymaddness/Procurement/files/3282442/debug.txt)

thailyn:
We need to be able to specify to an arbitrary ITransport implementation that we want to force refresh the stash tab.  This fixes a bug where, when using the CachedTransport, tabs would never be refreshed, since
forceRefresh would always be false.

Fixes #1029.

### DIFF
--- a/POEApi.Model/POEModel.cs
+++ b/POEApi.Model/POEModel.cs
@@ -144,7 +144,7 @@ namespace POEApi.Model
 
             onStashLoaded(POEEventState.BeforeEvent, index, -1);
 
-            using (var stream = Transport.GetStash(index, league, accountName, realm))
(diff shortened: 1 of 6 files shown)


aydjay:
Oh man - this mis my fuckup from when I introduced the xbox/ps4 support...

PR #1033, opened 2019-06-13
Clean up stash helper

### LINKED ISSUE #1031: Excluding tabs gives an unhandled error
Every time I choose to exclude a stash from Recipe results, I get a pop-up saying "there was an unhandled error" and Procurement crashes.

thailyn:
Better handle exceptions when fetching/building tab selector images, and moved some variables from function parameters to local parameters.

Fixes #1031.

### DIFF
--- a/Procurement/Utility/StashHelper.cs
+++ b/Procurement/Utility/StashHelper.cs
@@ -99,7 +99,23 @@ namespace Procurement.Utility
 
                     BitmapImage bitmapclone = (BitmapImage)bitmap.Clone();
                     bitmap = null;
-                    imageCache.Add(key, new CroppedBitmap(bitmapclone, new Int32Rect(0, offset, (int)bitmapclone.Width, 26)));
+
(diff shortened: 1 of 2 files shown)


Procurement/Utility/StashHelper.cs
```diff
@@ -99,22 +99,40 @@ private static System.Drawing.Bitmap buildImage(Tab tab, List<System.Drawing.Bit
 
                     BitmapImage bitmapclone = (BitmapImage)bitmap.Clone();
                     bitmap = null;
-                    imageCache.Add(key, new CroppedBitmap(bitmapclone, new Int32Rect(0, offset, (int)bitmapclone.Width, 26)));
+
+                    Int32Rect croppingRectangle = new Int32Rect();
+                    if (offset + 26 > bitmapclone.Height)
```
aydjay: 26 is a magic int which is being used later on, constant please :)

thailyn: Done.

PR #1036, opened 2019-06-16
Dynamically update a loot filter

### LINKED ISSUE #821: Recipe-specific dynamic loot-filter customizations
I use the [NeverSink](https://github.com/NeverSinkDev/NeverSink-Filter) loot filter (like probably most everyone who plays PoE), and in my local Procurement repo I've implemented updating the loot filter with some custom rules every time I recreate the recipes.  This mostly includes adding a rule to highlight normal items to finish partially-complete SameBaseType recipes, but I've also hard-coded in a list of base types to highlight for chancing specific uniques.

Right now it's pretty messy and has a bunch of stuff hard-coded, but I could work on cleaning it up and creating a PR if this is something that would be useful.  There would be a lot of customizations/options I would need to figure out how to do it in UI.  I also don't know if it would be useful to have support for modifying only one type of loot filter (it depends on searching for known strings in the NeverSink filter).

thailyn:
This PR adds limited support for updating loot filters after recreating the list of recipes.  There are two types of loot filters that can be added: items that are missing for partially-completed same-base-type recipes (that is, having a Normal, Magic, and Rare item of the same base type), with a variant for near-20% quality bases; and items that are chancing bases for builds in the current league.

The configuration for updating the item filters is managed by a Protocol Buffer `.proto` file.  Support has been added for Visual Studio to automatically compile the `.proto` file into C# classes that are virtually immediately recognized by Visual Studio's error-checking- and intelligence-generating systems.  Procurement reads and writes the actual configuration from and to a json file.

Closes #821.

### DIFF
--- a/POEApi.Model/ItemFilterConfig.json
+++ b/POEApi.Model/ItemFilterConfig.json
@@ -1,0 +1,83 @@
+﻿{
+  "fileConfig": [
+    {
+      "disabled": true,
+      "inputLocation": "<ITEM FILTER DIRECTORY>\\NeverSink's filter - 1-REGULAR.filter",
+      "outputLocation": "<ITEM FILTER DIRECTORY>\\[MOD] NeverSink's filter - 1-REGULAR.filter",
+      "filterRuleConfig": [
+        {
+          "ruleType": "CHANCING_BASES",
+          "name": "Chancing Bases",
+          "searchString": "#   [1008] Chancing items\r\n#------------------------------------",
+          "searchStringRelation": "AFTER",
+          "additionalConfig": {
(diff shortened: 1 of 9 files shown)


Stickymaddness:
Thanks for spending time on this, I'm curious what's the reason behind adding protobuf, I assume the main driver is the autogeneration and error checking?

thailyn:
Protobufs provide a strongly-typed and easy to understand data structure for storing user data, which can also be written to/read from JSON, with, indeed, automatic parsing and error checking.  The generated classes provide typical functions for handling the data. (Personally, I would strongly prefer to use Proto2 syntax instead of Proto3, but that is not supported by the official protobuf C# library, though it is, in my opinion, superior in several aspects.)

PR #1042, opened 2019-06-22
Automatically refresh used tab when log file changes

thailyn:
This PR accomplishes two things:
1) Add a class that monitors the client log file for changes, and reads new content when it has.  When it finds log messages indicating the player's character has changed locations, it triggers an event.
2) Add a handler for this event that refreshes all the user's used tabs.  This only happens at most ever two minutes.

Both of these behaviors are disabled by default.  The former is disabled when the config value `ClientLogFileLocation`, which should point to the log file, is missing or empty; or if the config value `EnableClientLogFileMonitoring` is missing or false.  The latter is disabled when the config value `EnableTabRefreshOnLocationChanged` is missing or false.

### DIFF
--- a/Procurement/Utility/ClientLogFileWatcher.cs
+++ b/Procurement/Utility/ClientLogFileWatcher.cs
@@ -71,6 +71,12 @@ namespace Procurement.Utility
             set;
         }
 
+        protected Regex LocationChangedRegex
+        {
+            get;
+            set;
+        }
+
         protected void Initialize()
         {
             if (!Settings.UserSettings.Keys.Contains("ClientLogFileLocation"))
@@ -87,6 +93,12 @@ namespace Procurement.Utility
             PollingTimer = new System.Timers.Timer();
             PollingTimer.Elapsed += (s, e) => { ReadClientLogFile(); };
             PollingTimer.Interval = 30000;  // 30 seconds
+
+            // Regex to catch lines formatted like:
+            //   2019/06/21 19:16:37 245842781 aa1 [INFO Client 19844] : You have entered Sunspire Hideout.
+            LocationChangedRegex = new Regex(
+                @"(\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}) (\d+) [^ .]* \[.*\] : You have entered (.*).$",
+                RegexOptions.Compiled);
         }
 
         internal void Start()
@@ -118,29 +130,27 @@ namespace Procurement.Utility
         {
             lock (Instance)
             {
-                var rx = new Regex(
-                    @"(\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}) (\d+) [^ .]* \[.*\] : You have entered (.*).$",
-                    RegexOptions.Compiled);
-
                 try
                 {
                     Stream stream = new FileStream(Settings.UserSettings["ClientLogFileLocation"], FileMode.Open,
                         FileAccess.Read, FileShare.ReadWrite);
                     using (var reader = new StreamReader(stream))
                     {
+                        // Quit early if the log file is no longer than the last time we read it.
                         if (reader.BaseStream.Length <= Instance.LastFileSizeSeen)
                         {
+                            Instance.LastFileSizeSeen = reader.BaseStream.Length;
                             return;
                         }
 
                         reader.BaseStream.Seek(Instance.LastFileSizeSeen, System.IO.SeekOrigin.Begin);
-                        string line;
+
                         DateTime eventTime = Instance.LastDateTimeSeen;
                         long eventTimestamp = Instance.LastTimestampSeen;
-                        string location;
+                        string line, location;
                         while ((line = reader.ReadLine()) != null)
                         {
-                            Match match = rx.Match(line);
+                            Match match = LocationChangedRegex.Match(line);
                             if (!match.Success)
                                 continue;
 
--- a/Procurement/Utility/ClientLogFileWatcher.cs
+++ b/Procurement/Utility/ClientLogFileWatcher.cs
@@ -9,6 +9,10 @@ namespace Procurement.Utility
 {
     class ClientLogFileWatcher
     {
+        private const string _clientLogFileLocationConfigName = "ClientLogFileLocation";
+        private const string _enableClientLogFileMonitoringConfigName = "EnableClientLogFileMonitoring";
+        private const int _clientLogFilePollingIntervalMilliseconds = 30000;  // 30 seconds
+
         private static ClientLogFileWatcher _instance;
         public static ClientLogFileWatcher Instance
         {
@@ -21,7 +25,7 @@ namespace Procurement.Utility
             }
         }
 
-        protected static System.IO.FileSystemWatcher FileWatcher
+        protected static FileSystemWatcher FileWatcher
         {
             get;
             private set;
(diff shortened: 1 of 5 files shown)


Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace Procurement.Utility
+{
+    public class ClientLogFileEventArgs : EventArgs
```
aydjay: Move this class into a seperate file

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            get
+            {
+                if (_instance == null)
+                    _instance = new ClientLogFileWatcher();
+
+                return _instance;
+            }
+        }
+
+        protected static System.IO.FileSystemWatcher FileWatcher
```
aydjay: System.IO is in the includes and the fully qualified path shouldn't be specified, this goes for the rest of the file.

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        protected Regex LocationChangedRegex
+        {
+            get;
+            set;
+        }
+
+        protected void Initialize()
+        {
+            if (!Settings.UserSettings.Keys.Contains("ClientLogFileLocation"))
```
aydjay: Move this magic string into a constant and update usage online 84

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            // Regex to catch lines formatted like:
+            //   2019/06/21 19:16:37 245842781 aa1 [INFO Client 19844] : You have entered Sunspire Hideout.
+            LocationChangedRegex = new Regex(
+                @"(\d{4}/\d{2}/\d{2} \d{2}:\d{2}:\d{2}) (\d+) [^ .]* \[.*\] : You have entered (.*).$",
+                RegexOptions.Compiled);
+        }
+
+        internal void Start()
+        {
+            if (!Settings.UserSettings.Keys.Contains("EnableClientLogFileMonitoring"))
```
aydjay: Same comment, move key to a constant

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            FileWatcher = new System.IO.FileSystemWatcher();
+            FileWatcher.Path = System.IO.Path.GetDirectoryName(fullFilePath);
+            FileWatcher.Filter = System.IO.Path.GetFileName(fullFilePath);
+            FileWatcher.NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.Size | NotifyFilters.LastAccess;
+
+            FileWatcher.Changed += OnFileChanged;
+
+            PollingTimer = new System.Timers.Timer();
+            PollingTimer.Elapsed += (s, e) => { ReadClientLogFile(); };
+            PollingTimer.Interval = 30000;  // 30 seconds
```
aydjay: Magic int. Constant needed.

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+        protected void Initialize()
+        {
+            if (!Settings.UserSettings.Keys.Contains("ClientLogFileLocation"))
+                return;
+            string fullFilePath = Settings.UserSettings["ClientLogFileLocation"];
+            if (string.IsNullOrWhiteSpace(fullFilePath))
+                return;
+
+            FileWatcher = new System.IO.FileSystemWatcher();
+            FileWatcher.Path = System.IO.Path.GetDirectoryName(fullFilePath);
```
aydjay: There is an assumption here that the fullFilePath provided by the user exists. Code could throw IO exceptions later on.

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            PollingTimer.Stop();
+        }
+
+        protected void ReadClientLogFile()
+        {
+            lock (Instance)
+            {
+                try
+                {
+                    Stream stream = new FileStream(Settings.UserSettings["ClientLogFileLocation"], FileMode.Open,
```
aydjay: Stream needs to be in a using, this is currently not being disposed. Use nested usings:

https://stackoverflow.com/questions/1329739/nested-using-statements-in-c-sharp#1329765

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+                        FileAccess.Read, FileShare.ReadWrite);
+                    using (var reader = new StreamReader(stream))
+                    {
+                        // Quit early if the log file is no longer than the last time we read it.
+                        if (reader.BaseStream.Length <= Instance.LastFileSizeSeen)
```
aydjay: I like this small performance tweak. 👍

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+                        string line, location;
+                        while ((line = reader.ReadLine()) != null)
+                        {
+                            Match match = LocationChangedRegex.Match(line);
+                            if (!match.Success)
+                                continue;
+
+                            eventTime = DateTime.ParseExact(match.Groups[1].Value, "yyyy/MM/dd HH:mm:ss",
+                                System.Globalization.CultureInfo.InvariantCulture);
+                            long.TryParse(match.Groups[2].Value, out eventTimestamp);
```
aydjay: Consume the try parse result and act accordingly

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        protected void Initialize()
+        {
+            if (!Settings.UserSettings.Keys.Contains("ClientLogFileLocation"))
+                return;
+            string fullFilePath = Settings.UserSettings["ClientLogFileLocation"];
+            if (string.IsNullOrWhiteSpace(fullFilePath))
+                return;
+
+            FileWatcher = new System.IO.FileSystemWatcher();
```
aydjay: I appreciate that all this code runs "Once" but just incase, null check the setup of both the FileWatcher here and the polling timer. If it does manage to hit the initialize method again it will leak the event handlers you are assigning.

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+            if (FileWatcher == null)
+                Initialize();
+
+            FileWatcher.EnableRaisingEvents = true;
+
+            if (!PollingTimer.Enabled)
+                PollingTimer.Start();
+        }
+
+        internal void Stop()
```
aydjay: Unsubscribe your event handlers to prevent memory leaks,

Procurement/Utility/ClientLogFileWatcher.cs
```diff
@@ -0,0 +1,191 @@
...
+
+            FileWatcher = new System.IO.FileSystemWatcher();
+            FileWatcher.Path = System.IO.Path.GetDirectoryName(fullFilePath);
+            FileWatcher.Filter = System.IO.Path.GetFileName(fullFilePath);
+            FileWatcher.NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.Size | NotifyFilters.LastAccess;
+
+            FileWatcher.Changed += OnFileChanged;
+
+            PollingTimer = new System.Timers.Timer();
+            PollingTimer.Elapsed += (s, e) => { ReadClientLogFile(); };
```
aydjay: Having an anonymous signature like this means you will be unable to unsubscribe from the event later on.

Procurement/ViewModel/LoginWindowViewModel.cs
```diff
@@ -185,6 +185,8 @@ public void Login(bool offline)
 
                 ApplicationState.SetDefaults();
 
+                ClientLogFileWatcher.Instance.Start();
```
aydjay: You have called start, but I can't see where you are calling stop. Is there an appropriate place to call stop?

Procurement/ViewModel/TabViewModel/StashViewModel.cs
```diff
@@ -153,6 +179,16 @@ public StashViewModel(StashView stashView)
                 currencyDistributionUsesCount = true;
             else
                 configuredOrbType = (OrbType)Enum.Parse(typeof(OrbType), currencyDistributionMetric);
+
+            if (Settings.UserSettings.Keys.Contains("EnableTabRefreshOnLocationChanged"))
```
aydjay: Magic string on the key

Procurement/Utility/ClientLogFileWatcher.cs
thailyn: The values set here indicate a pattern of files to watch, so the files don't have to exist here.  Later on we do catch `IOException`s when trying to read the actual log file.

Procurement/Utility/ClientLogFileWatcher.cs
thailyn: As per my other comment, I think this is ok, since the subscriber and publisher's lifetimes are the same, and does not need to be unsubscribed.  Please correct me if I am wrong.

Procurement/Utility/ClientLogFileWatcher.cs
thailyn: Done.

PR #1043, opened 2019-06-22
Set a custom API limit if specified in the config.

thailyn:
The HttpTransport will only allow a positive limit that is no larger than the real API limit.  A custom, smaller API limit will allow Procurement to play nicely with other applications that perform a known maximum number of API requests per minute.


POEApi.Transport/TaskThrottle.cs
```diff
@@ -55,7 +61,7 @@ public void StartTask()
                     }
 
                     RemvoeExpiredTasks();
```
aydjay: Appreciate it's not your typo, could you adjust please.

POEApi.Transport/TaskThrottle.cs
```diff
@@ -36,6 +36,12 @@ public TaskThrottle(TimeSpan windowSize, int windowLimit, int simultaneiousTasks
             SimultaneiousTasksLimit = simultaneiousTasksLimit;
```
aydjay: Typo

POEApi.Model/POEModel.cs
```diff
@@ -127,6 +127,22 @@ void instance_Throttled(object sender, ThottledEventArgs e)
 
         private ITransport GetTransport(string email, bool offline)
         {
+            int customRequestLimit;
+            if (Settings.UserSettings.Keys.Contains("ApiRequestLimit") &&
+                int.TryParse(Settings.UserSettings["ApiRequestLimit"], out customRequestLimit))
+            {
+                if (HttpTransport.AdjustThrottleWindowLimit(customRequestLimit))
```
aydjay: Can you change the message generation to a ternary please, more succinct.

POEApi.Transport/HttpTransport.cs
```diff
@@ -36,6 +36,8 @@ protected enum HttpMethod { GET, POST }
         private const string UpdateShopURL = @"https://www.pathofexile.com/forum/edit-thread/{0}";
         private const string BumpShopURL = @"https://www.pathofexile.com/forum/post-reply/{0}";
 
+        private const int _maximumWindowLimit = 42;
```
aydjay: Can this be brought in as an application setting?

POEApi.Transport/HttpTransport.cs
```diff
@@ -44,7 +46,21 @@ protected enum HttpMethod { GET, POST }
 
         public event ThottledEventHandler Throttled;
 
-        private static TaskThrottle _taskThrottle = new TaskThrottle(TimeSpan.FromMinutes(1), 42, 42);
+        private static TaskThrottle _taskThrottle = InitializeTaskThrottle();
+
+        protected static TaskThrottle InitializeTaskThrottle()
+        {
+            return new TaskThrottle(TimeSpan.FromMinutes(1), _maximumWindowLimit, _maximumWindowLimit);
```
aydjay: I can't tell just from the PR, isn't this just going to mindlessly overwrite the previously set task throttle?

POEApi.Transport/TaskThrottle.cs
thailyn: Done. (I rewrote the class, so, ultimately, it is my typo. :) )

POEApi.Transport/TaskThrottle.cs
thailyn: Done.

POEApi.Model/POEModel.cs
thailyn: Something like this?

POEApi.Transport/HttpTransport.cs
thailyn: I would rather not, for two reasons.
1) This value will likely very rarely -- if ever -- change, so the user will be unlikely to modify the setting.  He can also easily make Procurement work very poorly by increasing the value.
2) This class can't reference the `Settings` directly, since that would introduce a circular dependency between the projects.  Thus -- barring more substantial refactoring I'm not interested in doing now -- it would then have to be set after the static constructor finishes, potentially after it started handling requests.

POEApi.Transport/HttpTransport.cs
thailyn: It would, if it were called again.  As written, it is only ever invoked during `HttpTransport`'s static constructor.  It's also protected, so it can't be called by other classes, and `HttpTransport` itself doesn't call it.  In any case, I changed it back to not call a function, so it is less ambiguous.

Stickymaddness:
I'll take a look at this, and the other outsanding PR's as soon as I can.

Thanks for the continued contributions and effort 👍

PR #1044, opened 2019-06-23
Fix some memory leaks from recreating UserControls

thailyn:
**Note:** This PR slightly depends on the `log-file-monitor` branch (#1042), and includes it here in this PR as a base.  I'll rebase appropriately and make this PR ready for review when that PR has been merged.

Procurement recreates its stash and recipe views whenever they need updating.  This is inefficient and causes memory leaks (and this is readily apparent after automatically (and often) refreshing tabs).  This PR attempts to refresh those tabs instead of recreating them.  I believe I have addressed all the issues with reusing the same objects.

Note that the trading view is also recreated whenever tabs are refreshed; that is not addressed in this PR.

### DIFF
--- a/Procurement/Procurement.csproj
+++ b/Procurement/Procurement.csproj
@@ -109,6 +109,7 @@
     <Compile Include="Controls\FragmentStashTab.xaml.cs">
       <DependentUpon>FragmentStashTab.xaml</DependentUpon>
     </Compile>
+    <Compile Include="Utility\ClientLogFileWatcher.cs" />
     <Compile Include="ViewModel\Filters\ForumExport\FossilFilter.cs" />
     <Compile Include="ViewModel\Filters\ForumExport\FracturedItemFilter.cs" />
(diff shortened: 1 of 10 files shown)


nqrcqn:
I'm testing this, Currency Breakdown doesn't seem to change after refresh despite having more chaos than before

nqrcqn:
And it seems character inventories don't get refreshed, but regular stashes do.

thailyn:
Thank you so much for testing out these changes!  I just pushed another commit to this PR that should fix updating inventories; I had long suspected, but had not confirmed, that inventories were not being reported correctly in the "refresh all tabs" and similar windows, and that seems to indeed be the case (now fixed (though kinda hacky)).  I'll look into the currency breakdown not being updated properly shortly.  Thanks again!

aydjay:
I'll paw through this PR once you've marked it ready to merge, looks like we have a lot of outstanding PR's that need to be merged @Stickymaddness

PR #1057, opened 2019-09-04
Add divination cards from 3.8 (Blight)

nqrcqn:


nqrcqn:
Added the 3 new base item types introduced in Blight, and 4 missing fated uniques.

I think what's left is supporting the new Legion Scarabs, Oils and probably Blighted Maps

aydjay:
Great change. Looks good to me.

PR #1058, opened 2019-09-04
Add captured beasts to forum export

### LINKED ISSUE #1017: Add beasts to forum export
So we can sell those rare beasts and profit

nqrcqn:
closes #1017


aydjay:
Nice :)

PR #1059, opened 2019-09-04
Use white background for scarabs, incubators and emblems

### LINKED ISSUE #998: Show scarab titles with a gray background
At the moment, they are shown with a transparent background like unknown items. They can have the same gray background as some currencies and normal rarity items.

scarab

nqrcqn:
Same color used in-game and on trade. closes #998

### DIFF
--- a/Procurement/ViewModel/ItemHoverViewModelFactory.cs
+++ b/Procurement/ViewModel/ItemHoverViewModelFactory.cs
@@ -24,12 +24,7 @@ namespace Procurement.ViewModel
             if (stone != null)
                 r = stone.Rarity;
 
-            var vessel = item as DivineVessel;
-            if(vessel != null)
-                r = Rarity.Normal;
-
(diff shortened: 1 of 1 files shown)


aydjay:
Looks good to me, I would have prefered to just default to the currency view model if we don't explicitly define one.

Discussion here:

Are we sure it's safe to default to currency hover? What about if new types get added in the future?

_Originally posted by @Stickymaddness in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMTYzNDgzMTI1OnYy/pull_request_review_threads/discussion_

Stickymaddness:
@aydjay Your link seems to load a blank page.

nqrcqn:
How about returning currency if (item.MaxStackSize > 0) and return white (Rarity.Normal) for the rest.

PR #1060, opened 2019-09-05
Update stat filters

nqrcqn:
Filters now also support fractured, crafted and enchanted mods. Added new strings and some new filters.

I didn't add the following strings to IncreasedPhysicalDamageFilter and ManaRegenFilter since I don't know the difference between ExplicitModBase and (Or)StatFilter. Can I just convert them to OrStatFilter?

"increased Global Physical Damage"

"Mana Regenerated per second"
"as extra Mana Regeneration"

### DIFF
--- a/POEApi.Model/Gear.cs
+++ b/POEApi.Model/Gear.cs
@@ -11,6 +11,9 @@ namespace POEApi.Model
         public List<Socket> Sockets { get; set; }
         public List<SocketableItem> SocketedItems { get; set; }
         public List<string> Implicitmods { get; set; }
+        public List<string> Craftedmods { get; set; }
+        public List<string> Fracturedmods { get; set; }
+        public List<string> Enchantmods { get; set; }
         public List<Requirement> Requirements { get; set; }
         public GearType GearType { get; set; }
         public string BaseType { get; set; }
@@ -23,6 +26,9 @@ namespace POEApi.Model
             Explicitmods = item.ExplicitMods;
             SocketedItems = GetSocketedItems(item);
             Implicitmods = item.ImplicitMods;
+            Craftedmods = item.CraftedMods;
+            Fracturedmods = item.FracturedMods;
+            Enchantmods = item.EnchantMods;
             Requirements = ProxyMapper.GetRequirements(item.Requirements);
             ItemType = Model.ItemType.Gear;
             GearType = GearTypeFactory.GetType(this);
--- a/Procurement/ViewModel/Filters/ForumExport/ExplicitModBase.cs
+++ b/Procurement/ViewModel/Filters/ForumExport/ExplicitModBase.cs
@@ -14,12 +14,29 @@ namespace Procurement.ViewModel.Filters.ForumExport
 
         public bool Applicable(Item item)
         {
-            if (item.Explicitmods == null || !(item is Gear))
+            Gear gear = item as Gear;
+            if (gear == null)
                 return false;
 
-            foreach (var mod in item.Explicitmods)
-                if (mod.Contains(keyword))
-                    return true;
+            if (gear.Explicitmods != null)
+                foreach (var mod in gear.Explicitmods)
+                    if (mod.Contains(keyword))
+                        return true;
+
+            if (gear.Fracturedmods != null)
+                foreach (var mod in gear.Fracturedmods)
+                    if (mod.Contains(keyword))
+                        return true;
+
+            if (gear.Craftedmods != null)
+                foreach (var mod in gear.Craftedmods)
+                    if (mod.Contains(keyword))
+                        return true;
+
+            if (gear.Enchantmods != null)
+                foreach (var mod in gear.Enchantmods)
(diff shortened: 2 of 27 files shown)


POEApi.Model/Gear.cs
```diff
@@ -11,6 +11,9 @@ public class Gear : Item
         public List<Socket> Sockets { get; set; }
         public List<SocketableItem> SocketedItems { get; set; }
         public List<string> Implicitmods { get; set; }
+        public List<string> Craftedmods { get; set; }
```
thailyn: The `Item` base class already has properties for crafted mods and enchanted mods, so we shouldn't add them again here.  I'm thinking fractured mods should also be put in the `Item` class, but I'm not completely sure on that.

nqrcqn: I think fixed now

nqrcqn:
Talisman filter doesn't return the Talismans in my stash, also the Craftables filter returns very few results, may be fixed also.

nqrcqn:
To add new strings converted IncreasedPhysicalDamageFilter and ManaRegenFilter to use OrStatFilter

Can I add implicit mods to ExplicitModBase or was the goal there excluding implicit mods?

Procurement/ViewModel/Filters/ForumExport/FullBestiaryOrbFilter.cs
```diff
@@ -16,7 +16,7 @@ public string Keyword
         {
             get
             {
-                return "Full Bestiary Orb";
+                return "Captured Beasts";
```
thailyn: "Captured" beasts include those not just stored in orbs.  I used the name "Full Bestiary Orb" because that's essentially what GGG calls them -- see the file name used for the [icon](https://web.poecdn.com/image/Art/2DItems/Currency/BestiaryOrbFull.png): `BestiaryOrbFull.png`.

Procurement/ViewModel/Filters/ForumExport/GearSearchFilters.cs
```diff
@@ -14,6 +14,15 @@ class AmuletFilter : GearTypeFilter
         public AmuletFilter()
             : base(GearType.Amulet, "Amulets")
         { }
+        public override bool Applicable(Item item)
```
thailyn: Add a blank line before this one.

Procurement/ViewModel/Filters/ForumExport/ManaRegenFilter.cs
```diff
@@ -1,28 +1,17 @@
...
-        {
-            get { return "Items with increased Mana Regeneration Rate"; }
-        }
-
         public override FilterGroup Group
         {
             get { return FilterGroup.Default; }
         }
+        public ManaRegenFilter()
+            : base("Mana regen", "Items with increased Mana Regeneration Rate", "increased Mana Regeneration Rate", "Mana per second", "Mana Regenerated per second", "as extra Mana Regeneration")
```
thailyn: Split across multiple lines.

Procurement/ViewModel/Filters/ForumExport/ExplicitModBase.cs
```diff
@@ -14,12 +14,29 @@ public ExplicitModBase(string keyword)
...
+                foreach (var mod in gear.FracturedMods)
+                    if (mod.Contains(keyword))
+                        return true;
+
+            if (gear.CraftedMods != null)
+                foreach (var mod in gear.CraftedMods)
+                    if (mod.Contains(keyword))
+                        return true;
+
+            if (gear.EnchantMods != null)
```
thailyn: Technically, enchant mods are not explicit mods (and you can argue crafted and fractured mods aren't either).  However, I think the intent of this class (as it is being used) is to search the mods of an item.  As such, I could see also adding implicit mods to this class, if we rename the class.

Ideally, which mods are search should be a parameter to the value/property being searched for and can be customized by the user.  But that is a much larger change that is outside the scope of this PR.

Procurement/ViewModel/Filters/ForumExport/FullBestiaryOrbFilter.cs
nqrcqn: I used captured beasts because it's the same term used on trade, the other sounds more like an internal name

Procurement/ViewModel/Filters/ForumExport/ExplicitModBase.cs
nqrcqn: If there is interest, it will be easy to support customized searches on search box with my other PR, like name: type: mod: explicit: implicit: crafted: fractured: enchanted: flavour: property: requirement: description: prophecy:

nqrcqn: It seems to be added with only explicit mods in mind, so if we're going to add implicit mods, I think it's better to remove the file and use StatFilter instead

Issue #950, opened 2019-02-05
Support editing the currency list


nqrcqn:
While it would be better to support all possible currency tags the trade system supports, a simpler option would be to make it possible to edit the currency list that can be used as prices, for example I'd like to set exalted-shard as price for my chaos to buy shards without messing with the code manually.

Another option, custom currency tags might be typed directly from UI, if it's not going to mess up the current simplicity.

Stickymaddness:
This is a nice idea, the tags could be read from a config file, which could be edited to enable/disable more currency types.

PR #1067, opened 2019-09-14
Make currency type editable when pricing

nqrcqn:
related: #950


0x66656c6978:
How is this not merged yet.

PR #1093, opened 2020-03-08
TEMP: Switch to a patched CloudflareSolverRe package.

thailyn:
This package is a fork of the "official" CloudflareSolverRe, but it includes a fix for a Cloudflare change that was just introduced in the last day or so.

We should probably not merge this PR, and instead wait for the package we are using already to merge this patch.  In the meantime, however, this PR can provide a temporary working version of Procurement.

This PR also updates the Procurement projects to the .NET Framework 4.6.2 (from 4.5).  Newer versions of the CloudflareSolverRe package requires the .NET Framework 4.6.

**EDIT:** The latest version of this patch is from **2020.05.28**.  It can be downloaded [here](https://ci.appveyor.com/api/buildjobs/pcge9o9kjff03knt/artifacts/Procurement%2Fbin%2FProcurement.zip).  This download link is available for six months, until 2020.11.28.

PR #1114, opened 2020-06-17
Cloudflare problem workaround by nomis51

Brian-K-Smith:


Stickymaddness:
Sneaky workaround, but I've opted to merge #1093 as I can see GGG fixing this workaround in the future.

Issue #1148, opened 2022-12-22
[Error] Error downloading account name


chrlsfrmg:
Hi

[DebugInfo.log](https://github.com/Procurement-PoE/Procurement/files/10284139/DebugInfo.log)

Omga4000:
I've cloned this repository and (temporarily) fixed the issue.
Feel free to download the new version I've created from my own repository, it should be public.

https://github.com/Omga4000/Procurement/releases/tag/1.30.0

Let me know if that works for you.

chrlsfrmg:
![image](https://user-images.githubusercontent.com/67290803/209097867-f64b3807-ffd4-4c7c-9589-d853ad467e8a.png)

Omga4000:
@chrlsfrmg 
Your account name was not found.
Verified manually - it really wasn't found.

When was the last time you've logged into the game?
If it was before 11.08.2020, you will not be able to use it.
You can use this tool to check whether the API can find your account or not: https://poe.watch/Account

chrlsfrmg:
![image](https://user-images.githubusercontent.com/67290803/209098790-e165d5b8-9b55-45f3-8e42-8802cad689dd.png)
?

chrlsfrmg:
> @chrlsfrmg O nome da sua conta não foi encontrado. Verificado manualmente - realmente não foi encontrado.
> 
> Quando foi a última vez que você entrou no jogo? Se foi antes de 11.08.2020, você não poderá usá-lo. Você pode usar esta ferramenta para verificar se a API pode encontrar sua conta ou não: https://poe.watch/Account

I'm playing now!

chrlsfrmg:
![image](https://user-images.githubusercontent.com/67290803/209099476-9df8bd94-8906-4dc5-a75d-84f9391d0aab.png)

Why are you giving this? I'm playing right now, I've been playing for days now.

Omga4000:
@chrlsfrmg 
False positive, I assume. 
Did you ever scan 1.29.1 to see if the result is the same?
You can fully see all of my changes and see I made no significant ones, definitely not something that will trigger a virus alter, unless the original Procurement also triggered it - GitHub doesn't hide anything.

Regarding why this is not working for you, I cannot say.
I gave you this website because the changes I've made are based off of poe-watch's API and if they cannot find you, it won't work.
You'll need to resolve it and have that website recognize you before you can use my changes.

chrlsfrmg:
[...]

Issue #1147, opened 2022-12-16
hollowaycs


Hollowaycs:
im jsut gonna copy and past what was in the DebugInfo folder. please resolve as sonnas u can

[16-12-2022 17:35] System.Exception: Error downloading account name, details logged to DebugInfo.log. Please open a ticket at https://github.com/Stickymaddness/Procurement/issues and include your DebugInfo.log
   at POEApi.Model.POEModel.GetAccountName(String realm)
   at Procurement.ViewModel.LoginWindowViewModel.<>c__DisplayClass23_0.<Login>b__0()
   at System.Threading.Tasks.Task.Execute()

Omga4000:
I've cloned this repository and (temporarily) fixed the issue.
Feel free to download the new version I've created from my own repository, it should be public.

https://github.com/Omga4000/Procurement/releases/tag/1.30.0

Let me know if that works for you.

Omga4000:
Please try the latest release. It contains a new fix which should work permanently.
https://github.com/Omga4000/Procurement/releases/

thailyn:
Duplicate of #1140 (as far as I can tell from the info provided).

Issue #1146, opened 2022-12-11
error !!can someone help me?


DarkyUzzy:
Procurement 1.29.1 Initialized.
                                                          
Authenticating DarkyUzzy                                                                  [OK]
Fetching account name...                                                                  

[Er
[DebugInfo.log](https://github.com/Procurement-PoE/Procurement/files/10201764/DebugInfo.log)
ror] Error downloading account name, details logged to DebugInfo.log. Please open a ticket at https://github.com/Stickymaddness/Procurement/issues and include your DebugInfo.log

chrlsfrmg:
managed to solve?

Omga4000:
I've cloned this repository and (temporarily) fixed the issue.
Feel free to download the new version I've created from my own repository, it should be public.

https://github.com/Omga4000/Procurement/releases/tag/1.30.0

Let me know if that works for you.

DarkyUzzy:
> managed to solve?

no.
i tried the new one but still down...

Omga4000:
@DarkyUzzy 
In order for this fix to work, your account must be recognized by Poe Watch.
To check if your account is recognized, please use the following tool:
https://poe.watch/Account

Omga4000:
@DarkyUzzy 

Please try the latest release. It contains a new fix which should work permanently.
https://github.com/Omga4000/Procurement/releases/

thailyn:
Duplicate of #1140.

Issue #1144, opened 2022-11-07
Error downloading account name plis


dequer666:
[DebugInfo.log](https://github.com/Procurement-PoE/Procurement/files/9956195/DebugInfo.log)

chrlsfrmg:
managed to solve?

Omga4000:
I've cloned this repository and (temporarily) fixed the issue.
Feel free to download the new version I've created from my own repository, it should be public.

https://github.com/Omga4000/Procurement/releases/tag/1.30.0

Let me know if that works for you.

Omga4000:
Please try the latest release. It contains a new fix which should work permanently.
https://github.com/Omga4000/Procurement/releases/

thailyn:
Duplicate of #1140.

PR #1149, opened 2022-12-22
[#1140,#1141,#1142,#1144,#1146,#1147,#1148] Temporary fix for "Error downloading account name"

Omga4000:


thailyn:
Thanks for sending a pull request to fix the issue.  In the future, you can rebase away or otherwise get rid of initial changes that are no longer relevant (i.e., the switch to using `https://api.poe.watch/char` temporarily), and don't include version number changes in feature pull requests.

